PR #4, opened 2017-04-28
Add support for Integer Dynamic Property Value Type.

lordstyx:
This pull request goes together with the one I made for the vc-module-core repository. Note that this change can't be built without it.

Since the data model already supports Integer as value type for dynamic properties, I propose supporting it in the front-end as well.


tatarincev:
Could you accept [licence agreement ](https://virtocommerce.com/contribute-agreement) before I  merge your PR? And could you make PR to dev branch?

lordstyx:
I've accepted the license agreement and made a new PR to the dev branch.

PR #10, opened 2017-06-20
Ability to view/edit GTIN property on product blade

ksavosteev:


ksavosteev:
#872

tatarincev:
I think better position for Gtin number will be as it displayed on image 
image

PR #13, opened 2017-07-28
Expose DictionaryProperty Alias in front-end

lordstyx:
The Alias functionality is already supported in the backend. This is just an exposure of it through front-end.


ksavosteev:
@lordstyx Thanks for the commit. Added the same functionality for non multi language props.

PR #14, opened 2017-07-28
Introduce a NearRealtimeIndexing feature, Fix index deletes for CatalogRepositoryImpl

lordstyx:
The NearRealtimeIndexing will only kick in when automatic indexing is enabled and indexing by job is disabled (see Settings -> Search -> General).
Note that to enable this, you must have a module that implements ISearchProdiver.

Used factories to make sure there's no dependency on any ISearchProvider implementation in Catalog's Module Initialize.

CatalogRepositoryImpl now removes documents from the index after manually deleting them from the DB.
Not sure if the EF context should be updated there as well, can you verify?


tatarincev:
My opinion is real time indexation should be implemented on different layer, not in persistent layer as you made. Currently we do not have any of events for catalog model which you could use for different tasks, one of it is real time indexation.  But we have some plans to add some events in the catalog module.

PR #46, opened 2017-10-02
Product and Category Details to metaform

ksavosteev:


tatarincev:
It is necessary to make these meta-forms expandable from other modules, for this need to define some special service for resolving these forms meta-data for future extensions.
See how it made in customer 
https://github.com/VirtoCommerce/vc-samples/blob/master/MemberExtensionSampleModule/Scripts/extMembers.js#L14

tatarincev:
@ksavosteev  Could you leave meat-form fields registration in individual blades controllers? Because catalog.js  It looks very thick:)

PR #47, opened 2017-10-09
Updated nuget packages VirtoCommerce.Domain and VirtoCommerce.CoreModule.Data to version 2.24.13

megafreeman:


tatarincev:
You forgot to update dependency in `module.manifest` file

PR #53, opened 2017-10-18
Can include variations to Associations

eMazeika:
Catalog items search fixed to include variations;
Now manager can include variations to Associations.


VirtoCommerce.CatalogModule.Web/Scripts/blades/common/catalog-items-select.js
```diff
@@ -27,6 +27,7 @@
                     categoryId: blade.categoryId,
                     keyword: filter.keyword,
                     responseGroup: blade.responseGroup || 'withCategories, withProducts',
+                    searchInVariations: true,
```
tatarincev: Can we do it optionally, to be able to control this behavior where this control uses?

Issue #30, opened 2017-09-05
Configurable storage blob provider path for store product images


OlegoO:
We should use folder path  
[CATALOG_CODE]/[PRODUCT_SKU]
when upload image to blob - File and Azure providers. 

@tatarincev commented on [Thu Jun 16 2016](https://github.com/VirtoCommerce/vc-platform/issues/512)

Currently new images uploaded to  storage path with product.code name. need to create some special setting which will be used for generate  storage path.
Some like this: {catalog}_{product.code}

n2pro:
Extended path for storing an image for a product
[catalog]/[category.code][product.code] or [catalog]/[product.code]

tatarincev:
Released https://github.com/VirtoCommerce/vc-module-catalog/releases/tag/v2.20.6

PR #68, opened 2017-11-03
Extension storage path to product image #30

n2pro:


VirtoCommerce.CatalogModule.Web/Scripts/resources/categories.js
```diff
@@ -1,7 +1,8 @@
 ﻿angular.module('virtoCommerce.catalogModule')
 .factory('virtoCommerce.catalogModule.categories', ['$resource', function ($resource) {
 
-	return $resource('api/catalog/categories/:id', { categoryId: '@Id' }, {
+    return $resource('api/catalog/categories/:id', { categoryId: '@Id' }, {
```
tatarincev: This method declaration is redundant. 
Follow construction will work without this definition `categories.get({ id: blade.currentEntityId }`

VirtoCommerce.CatalogModule.Web/Scripts/blades/images-add.js
```diff
@@ -2,8 +2,8 @@
 .controller('virtoCommerce.catalogModule.imagesAddController',
     ['$scope', '$filter', '$translate', 'FileUploader', 'platformWebApp.dialogService',
         'platformWebApp.bladeNavigationService', 'platformWebApp.authService',
-        'platformWebApp.assets.api', 'virtoCommerce.catalogModule.imageTools', 'platformWebApp.settings',
-        function ($scope, $filter, $translate, FileUploader, dialogService, bladeNavigationService, authService, assets, imageTools, settings) {
+        'platformWebApp.assets.api', 'virtoCommerce.catalogModule.imageTools', 'platformWebApp.settings', 'virtoCommerce.catalogModule.categories',
+        function ($scope, $filter, $translate, FileUploader, dialogService, bladeNavigationService, authService, assets, imageTools, settings, categories) {
```
tatarincev: Redundant `categories` injection is not used anywhere

VirtoCommerce.CatalogModule.Web/Scripts/blades/images-add.js
```diff
@@ -39,7 +40,7 @@
                         removeAfterUpload: true
                     });
 
-                    uploader.url = getImageUrl(item.code, blade.imageType).relative;
+                    uploader.url = getImageUrl(item.code, blade.imageType, item.category).relative;
```
tatarincev: This controller reused in other places. and `item` here might be a category type.
I this is a not correct way to try get the storage path from element here, the better solution would be to pass a storage folder path as a special parameter to this controller.

VirtoCommerce.CatalogModule.Web/Scripts/resources/categories.js
n2pro: remove this get

VirtoCommerce.CatalogModule.Web/Scripts/blades/category-detail.js
```diff
@@ -34,13 +35,18 @@
         blade.title = data.name;
         blade.isLoading = false;
         blade.securityScopes = data.securityScopes;
+        blade.currentEntity.folderPath = (data.catalogId ? (getShortCatalogId(data.catalogId) + '/') : '') + data.code;
```
tatarincev: This line leads to forcing category modification it is really the bug. The folder path initialization code should be moved to the CategoryImageWidget controller.

VirtoCommerce.CatalogModule.Web/Scripts/blades/item-detail.js
```diff
@@ -30,6 +29,10 @@
             blade.currentEntity = blade.item;
             blade.origItem = data;
             blade.isLoading = false;
+
+            //set folderPath
```
tatarincev: This line leads to forcing Item modification it is really the bug. The folder path initialization code should be moved to the ItemImageWidget controller.

Issue #67, opened 2017-11-02
Add StartDate and EndDate to product UI


Woland2k:
Currently index does add StartDate and EndDate for product, however there is no UI for those properties.

tatarincev:
released https://github.com/VirtoCommerce/vc-module-catalog/releases/tag/v2.20.5

Woland2k:
We should allow entering time as well as date for this.

tatarincev:
We have another issue related to this requirement 
https://github.com/VirtoCommerce/vc-platform/issues/1204

PR #69, opened 2017-11-07
Add StartDate and EndDate to product UI #67

n2pro:
Add view  (save, update) StartDate and EndDate to product without functionality


VirtoCommerce.CatalogModule.Data/Model/ItemEntity.cs
```diff
@@ -344,6 +344,8 @@ public virtual void Patch(ItemEntity target)
             target.ShippingType = this.ShippingType;
             target.Priority = this.Priority;
             target.ParentId = this.ParentId;
+            target.StartDate = this.StartDate;
+            target.EndDate = this.EndDate;
```
tatarincev: You forgot to add these fields to the Patch method, without this change, modification the values of this field will not work.

n2pro: this.StartDate and this.EndDate they have already been added

PR #71, opened 2017-11-14
Map alias in property value converter.

vanzegbroecke:
Return aliases through the api.


asvishnyakov:
@vanzegbroecke Please, add your email vanzegbroecke@delawareconsulting.com to your GitHub account, because otherwise GitHub can't understand what this commit made by you.

PR #76, opened 2017-11-23
Optionally use category codes and product code in outline instead of ids

WimVergouwe:
Allow using category and products codes to generate outlines.


WimVergouwe:
@tatarincev, I agree, see commit.

PR #82, opened 2017-12-07
Replacing datepicker for product of datetime picker #1204

n2pro:


n2pro:
Add date time picker without modify other modules

Issue #84, opened 2017-12-08
Add Geo-point datatype into Product's properties


OlegoO:
@OlegoO commented on [Wed Jan 18 2017](https://github.com/VirtoCommerce/vc-platform/issues/703)

As Client, I want to display and search products on map by geo-point.

### References
A point representing a geographic location on the globe:
https://www.elastic.co/guide/en/elasticsearch/guide/master/geopoints.html
https://docs.microsoft.com/en-us/rest/api/searchservice/supported-data-types

### Tasks
- [ ] Add ability to add a new product property as Geo-point. 
- [ ] Elastic Search should index it as geo-point https://www.elastic.co/guide/en/elasticsearch/reference/current/geo-point.html
- [ ] Search service should provide ability to find nearest products  by geo-point.
- [ ] Search service should provide ability to order by distance from geo-point.

### Example
Geo-point as a string: "41.12,-71.34"

tatarincev:
Released https://github.com/VirtoCommerce/vc-module-catalog/releases/tag/v2.21.0

PR #91, opened 2017-12-15
Add geo point property #84

n2pro:


VirtoCommerce.CatalogModule.Web/Scripts/directives/customPropertyValidation.js
```diff
@@ -0,0 +1,23 @@
+﻿angular.module('virtoCommerce.catalogModule')
+    .directive('geoPointValidation', function () {
```
tatarincev: Our custom directives names should start with prefix `va-`

PR #105, opened 2018-01-18
Fixed product search within Categories.

UnreliableWitness:
The product search doesn't work anymore after the change of Outlines to use Codes instead of Id's.
When searching, the id was still used to query instead of the outline.

* Assured outlines are fetched and sent to browser
* Assured outline is sent back on search
* Fixed issue where Catalog was added twice in SearchCriteriaExtensions.GetOutlines

### DIFF
--- a/VirtoCommerce.CatalogModule.Data/Services/CatalogSearchServiceDecorator.cs
+++ b/VirtoCommerce.CatalogModule.Data/Services/CatalogSearchServiceDecorator.cs
@@ -54,7 +54,7 @@ namespace VirtoCommerce.CatalogModule.Data.Services
                 serviceCriteria.ObjectType = KnownDocumentTypes.Product;
                 serviceCriteria.SearchPhrase = criteria.Keyword;
                 serviceCriteria.CatalogId = criteria.CatalogId;
-                serviceCriteria.Outline = criteria.Outline;
+                serviceCriteria.Outline = criteria.Outline ?? criteria.CategoryId;
                 serviceCriteria.WithHidden = criteria.WithHidden;
                 serviceCriteria.Skip = criteria.Skip;
                 serviceCriteria.Take = criteria.Take;
--- a/VirtoCommerce.CatalogModule.Data/Services/CatalogSearchServiceDecorator.cs
+++ b/VirtoCommerce.CatalogModule.Data/Services/CatalogSearchServiceDecorator.cs
@@ -54,7 +54,7 @@ namespace VirtoCommerce.CatalogModule.Data.Services
                 serviceCriteria.ObjectType = KnownDocumentTypes.Product;
                 serviceCriteria.SearchPhrase = criteria.Keyword;
                 serviceCriteria.CatalogId = criteria.CatalogId;
-                serviceCriteria.Outline = criteria.CategoryId;
+                serviceCriteria.Outline = criteria.Outline;
                 serviceCriteria.WithHidden = criteria.WithHidden;
                 serviceCriteria.Skip = criteria.Skip;
                 serviceCriteria.Take = criteria.Take;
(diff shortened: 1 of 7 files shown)


VirtoCommerce.CatalogModule.Web.Core/Converters/CategoryConverter.cs
```diff
@@ -39,7 +39,7 @@ public static webModel.Category ToWebModel(this moduleModel.Category category, I
             var parents = new List<moduleModel.Category>();
             if (category.Parents != null)
             {
-                retVal.Outline = string.Join("/", category.Parents.Select(x => x.Id));
+                retVal.Outline = category.Outlines.FirstOrDefault().ToString();
```
artem-dudarev: Previous version always returned physical outline, and new version can return virtual outline - this is a breaking change.
Need to replace Outlines.FirstOrDefault() with the code, that selects the physical outline.

VirtoCommerce.CatalogModule.Data/Services/CatalogSearchServiceDecorator.cs
```diff
@@ -54,7 +54,7 @@ public virtual SearchResult Search(SearchCriteria criteria)
                 serviceCriteria.ObjectType = KnownDocumentTypes.Product;
                 serviceCriteria.SearchPhrase = criteria.Keyword;
                 serviceCriteria.CatalogId = criteria.CatalogId;
-                serviceCriteria.Outline = criteria.CategoryId;
+                serviceCriteria.Outline = criteria.Outline;
```
artem-dudarev: Need to add backward compatibility with existing clients that still use CategoryId:
`serviceCriteria.Outline = criteria.Outline ?? criteria.CategoryId`

UnreliableWitness:
Thanks for reviewing! I have accepted the CLA but the check is still unconfirmed. Do I have to take other actions to enable the merge to dev?

tatarincev:
At first, I think that you need to fix all the remarks related to backward compatibility breaking-changes.

UnreliableWitness:
Hey @tatarincev , I've made the changes as requested. Can you review please?

VirtoCommerce.CatalogModule.Web.Core/Converters/CategoryConverter.cs
```diff
@@ -39,7 +39,8 @@ public static webModel.Category ToWebModel(this moduleModel.Category category, I
             var parents = new List<moduleModel.Category>();
             if (category.Parents != null)
             {
-                retVal.Outline = string.Join("/", category.Parents.Select(x => x.Id));
+                retVal.Outline = OutlineHelper.EnsurePhysicalOutline(category.Outlines, category.CatalogId)
```
artem-dudarev: This code will add physical catalog ID to every outline, even if it's not physical, and take the first one.
Instead, it should find physical outline and return its string representation.

VirtoCommerce.CatalogModule.Web.Core/Converters/OutlineHelper.cs
```diff
@@ -0,0 +1,37 @@
...
+                }
+            }
+
+            return retVal.ToArray();
+        }
+
+        public static IEnumerable<string> EnsurePhysicalOutline(ICollection<Outline> outlines, string catalogId)
+        {
+            var x = outlines.Select(ol => ol.ToString()).ToArray();
+            return EnsurePhysicalOutline(x, catalogId);
```
artem-dudarev: This is inefficient. You are joining outline parts into a single string, and then splitting it to parts again.

Also outline.ToString() is intended basically for debugging/testing. Its value can contain `*` indicating that outline part has virtual parent. Such strings should not be returned from API.

PR #114, opened 2018-02-05
getting module connection string changed

MalikJuly:


VirtoCommerce.CatalogModule.Web/Module.cs
```diff
@@ -33,7 +33,7 @@ namespace VirtoCommerce.CatalogModule.Web
 {
     public class Module : ModuleBase, ISupportExportImportModule
     {
-        private static readonly string _connectionString = ConfigurationHelper.GetNonEmptyConnectionStringValue("VirtoCommerce");
+        private readonly string _connectionStringName = ConfigurationHelper.GetConnectionStringValue("VirtoCommerce.Catalog") ?? ConfigurationHelper.GetConnectionStringValue("VirtoCommerce");
```
artem-dudarev: It's not a name. It's a connection string itself.

VirtoCommerce.CatalogModule.Web/web.config
```diff
@@ -39,7 +39,7 @@
         <bindingRedirect oldVersion="0.0.0.0-1.3.0.0" newVersion="1.3.0.0" />
       </dependentAssembly>
     </assemblyBinding>
-  </runtime>
+  </runtime>
```
artem-dudarev: We don't need this space at the end of the line.

Issue #92, opened 2017-12-18
There is no validation when adding a new dictionary type property


n2pro:
When adding a new property for integer and other properties no validation adding values to the dictionary
![capture](https://user-images.githubusercontent.com/13115017/34095778-75c4dfba-e3db-11e7-80c4-de723e10ef2a.JPG)

tatarincev:
The blade for managing the property dictionary
![image](https://user-images.githubusercontent.com/7566324/36794353-c1be3a3c-1ca8-11e8-8a6d-88ad2e0a0cd8.png)

tatarincev:
Released https://github.com/VirtoCommerce/vc-module-catalog/releases/tag/v2.23.0

PR #115, opened 2018-02-06
#92 used va-property2 to validate values in each editable fields

sergfed:
#92


VirtoCommerce.CatalogModule.Web/Scripts/blades/property-detail.js
```diff
@@ -71,6 +71,7 @@
             }
 
             blade.currentEntity = angular.copy(data);
+            blade.currentEntity.isValid = true;
```
tatarincev: This line makes currentEntity always dirty, therefore save button always enabled even if property unchanged.
use '$" as leading char for property names if you want to do not use them for equality comparison. 
`blade.currentEntity.$isValid` as example

sergfed:
When we add a multilanguage value there is no visual notification that the value is empty. The "va-property2" doesn't show the notification (there is no notifications at all that the value is empty). Should I implement the notification separately for this case?

PR #120, opened 2018-02-13
119 fixed bug with negative quantity & height fields

MalikJuly:


tatarincev:
Released https://github.com/VirtoCommerce/vc-module-catalog/pull/120

PR #121, opened 2018-02-15
Negative quantities etc.

MalikJuly:
Fixed validation of value of min & max quantities, weight, height, width, length fields

### DIFF
--- a/VirtoCommerce.CatalogModule.Web/Scripts/blades/item-dimensions.tpl.html
+++ b/VirtoCommerce.CatalogModule.Web/Scripts/blades/item-dimensions.tpl.html
@@ -33,7 +33,7 @@
                             <div class="form-group">
                                 <label class="form-label">{{ 'catalog.blades.item-detail.labels.weight' | translate }}</label>
                                 <div class="form-input __number">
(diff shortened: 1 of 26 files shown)


VirtoCommerce.CatalogModule.Web/Scripts/blades/item-dimensions.tpl.html
```diff
@@ -56,23 +56,23 @@
...
+                                  <input num-type="float" min="0.00" ng-pattern="/^\d+$/" smart-float ng-model="blade.currentEntity.height" placeholder="N/A" />
                                 </div>
                             </div>
                         </div>
                         <div class="column">
                             <div class="form-group">
                                 <label class="form-label">{{'catalog.widgets.itemDimensions.width' | translate}}</label>
                                 <div class="form-input __number">
-                                    <input smart-float num-type="float" ng-model="blade.currentEntity.width" placeholder="N/A" />
+                                    <input smart-float min="0.00" num-type="float" ng-pattern="/^\d+$/" ng-model="blade.currentEntity.width" placeholder="N/A" />
```
tatarincev: Try to use this markup ```input smart-float min="0.00" num-type="integer">```
instead of write regexp in each tag, because is a bad design to duplicate regexp expression in **html** elements

PR #125, opened 2018-02-27
property-dictionary blade rewritten

sergfed:
#92 the blade 'property-dictionary' was fully rewritten.
There is one problem remains. There is no focus on an input element when a user selects an item in the dictionary value list (multilanguage mode).

### DIFF
--- a/VirtoCommerce.CatalogModule.Web/Content/css/catalog.css
+++ b/VirtoCommerce.CatalogModule.Web/Content/css/catalog.css
@@ -251,3 +251,47 @@
 .capitalizeText {
     text-transform: capitalize;
 }
+
+
+
+.vc-catalog .property-dictionary-new {
+    height: auto !important;
+}
+.vc-catalog .property-dictionary-new .__checkbox {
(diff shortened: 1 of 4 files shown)


VirtoCommerce.CatalogModule.Web/Content/css/catalog.css
```diff
@@ -250,4 +250,48 @@
...
+}
+.vc-catalog .property-dictionary-list .form .table .table-col {
+    padding-top: 0 !important;
+}
+.vc-catalog .property-dictionary-list .form .form-input {
+    margin: 10px 0 0 !important;
+}
+.vc-catalog .property-dictionary-list .form .form-group {
+    margin: 0 !important;
+}
```
asvishnyakov: Why we need this? Will cause problems if we will try to change this styles in platform: we will need to fix it in catalog module too, so platform start depend on module, what's wrong.

tatarincev:
Propery dictionary wireframe
image

tatarincev:
@Tilesto is working on another implementation

PR #126, opened 2018-02-27
Added required attribute to Locale property of EditorialReviewEntity

MalikJuly:


tatarincev:
I think the more right solution will to make Language as optional for product editorial reviews. Thus we obtain a default|fallback data for usage on client-side.

PR #131, opened 2018-03-14
Added possibility to use thumbnails instead of full size images for products listing

MalikJuly:


tatarincev:
Use similar to this solution for this task
https://stackoverflow.com/questions/27896833/filter-file-name-with-angularjs
And the result should be like this ` ng-src={{row.entity.imageUrl | thumbnail 64x64}}`

MalikJuly:
ок

PR #150, opened 2018-07-10
changed blade search panel to general style

Tilesto:
image


VirtoCommerce.CatalogModule.Web/Content/css/catalog.css
```diff
@@ -250,4 +250,13 @@
 }
 .capitalizeText {
     text-transform: capitalize;
-}
\ No newline at end of file
+}
+
+vc-item-search {
```
yecli: I think there could be something more elegant. Probably you can discuss it with @anivirtoway.

PR #158, opened 2018-08-16
Catalog should not allow to pass multiple product descriptions with s…

anton-abyzov:
…ame both language & type#152


VirtoCommerce.CatalogModule.Web/Scripts/blades/editorialReview-detail.js
```diff
@@ -109,7 +109,18 @@
         };
 
         function canSave() {
-            return isDirty() && formScope && formScope.$valid;
+            var thisBlade = blade;
```
tatarincev: I would suggest moving this logic to the new function `isValid()`
with follow code
```
function isValid(){
var result = formScope && formScope.$valid;
//Check  for duplications for new editorial  reviews
if(result && !blade.currentEntity.Id)
{
  result =  _.any(blade.item.reviews, function(x) {
     return x.reviewType === blade.currentEntity.reviewType && x.languageCode === blade.currentEntity.languageCode;
 });
}
return result;
}
```

and call isValid() in canSave function

Issue #162, opened 2018-08-28
Allow to search by product type in indexed search


asvishnyakov:
Currently only criteria for database search has ProductType property. Indexed search isn't support it

tatarincev:
Released https://github.com/VirtoCommerce/vc-module-catalog/releases/tag/v2.22.1

PR #166, opened 2018-09-07
Search by product type in indexed search impl. #162

eMazeika:
Search by product type in indexed search impl. #162


VirtoCommerce.CatalogModule.Data/Search/ProductSearchRequestBuilder.cs
```diff
@@ -153,6 +153,11 @@ protected virtual IList<IFilter> GetPermanentFilters(ProductSearchCriteria crite
                 result.Add(FiltersHelper.CreateTermFilter("__type", criteria.ClassTypes));
             }
 
+            if (!string.IsNullOrEmpty(criteria.ProductType))
```
tatarincev: You can use this property for this purpose https://github.com/VirtoCommerce/vc-module-core/blob/master/VirtoCommerce.Domain/Catalog/Model/Search/ProductSearchCriteria.cs#L24

PR #173, opened 2018-09-21
Support for custom isItemSelectable function

artem-dudarev:
This will allow to provide a function that will determine whether the catalog item can be selected


artem-dudarev:
Oops, sorry

VirtoCommerce.CatalogModule.Web/Scripts/blades/common/catalog-items-select.js
```diff
@@ -108,12 +108,12 @@
...
 
         var newBlade = {
-            id: 'CatalogItemsSelect',
+            id: blade.id,
             breadcrumbs: blade.breadcrumbs,
             catalogId: blade.catalogId,
             catalog: blade.catalog,
-            controller: 'virtoCommerce.catalogModule.catalogItemSelectController',
-            template: 'Modules/$(VirtoCommerce.Catalog)/Scripts/blades/common/catalog-items-select.tpl.html',
+            controller: blade.controller,
```
tatarincev: Are you sure what these changes won't affect other places where this control used?

artem-dudarev: I'm pretty sure it won't affect anything, but I can remove it, if you wish.

eMazeika: looks like a nice code clean-up, just double-check to be sure

Issue #174, opened 2018-09-26
Switch the selector ui used to select the properties dictionary entries to dynamically data loading


tatarincev:
![image](https://user-images.githubusercontent.com/7566324/46073880-68090480-c186-11e8-8b9a-a4c3af2a0c1c.png)

The **property2** directive which used as the primary component for edit property values
https://github.com/VirtoCommerce/vc-module-catalog/blob/master/VirtoCommerce.CatalogModule.Web/Scripts/directives/property2.js
https://github.com/VirtoCommerce/vc-module-catalog/blob/master/VirtoCommerce.CatalogModule.Web/Scripts/directives/property2.tpl.html#L158

The single place where this directive  used
https://github.com/VirtoCommerce/vc-module-catalog/blob/master/VirtoCommerce.CatalogModule.Web/Scripts/blades/property-list.tpl.html#L17

Need to switch for dynamic data loading  all the following property templates
ShortText-dictionary.html
ShortText-dictionary-multivalue-multilang.html
ShortText-dictionary-multilang.html
Numeric-dictionary.html
Numeric-dictionary-multivalue-multilang.html
Numeric-dictionary-multilang.html
 

See as for example this commit
https://github.com/VirtoCommerce/vc-module-customer/commit/e15099d010004f61c1002193981709d6b8b5a528#diff-319042bdd209e32ac44ae46921496540

tatarincev:
Released https://github.com/VirtoCommerce/vc-module-catalog/releases/tag/v2.23.4

PR #175, opened 2018-10-02
174: added infinite scrolling for dictionary property values

avas:
Changes made in `va-platform2` directive for #174:
* `getPropValues` function: added `countToSkip` and `countToTake` parameters. They are used for paged parameter loading.
* Added `loadNextDictionaryValues` method which is called by scrolling dictionary value list down. It loads next page of dictionary values (if the backend has more values to send). It is also used internally by `loadDictionaryValues` method which loads first page of values.
* Added optional `page-size` attribute that is used to adjust page size. If it is not specified, `va-property2` will use `50` as a fallback value.
* Modified templates for lists of dictionary values: replaced hard-coded English placeholders by translatable resource strings from `??.VirtoCommerce.Common.json` resources (located in `VirtoCommerce.Platform`).

### DIFF
--- a/VirtoCommerce.CatalogModule.Web/Scripts/directives/property2.js
+++ b/VirtoCommerce.CatalogModule.Web/Scripts/directives/property2.js
@@ -128,11 +128,10 @@ angular.module('virtoCommerce.catalogModule')
 
                 return scope.getPropValues()(scope.currentEntity.id, '', countToSkip, countToTake).then(function (result) {
                     populateDictionaryValues(result.results);
+                    $select.page++;
 
                     // If there are more items to display, let's prepare to handle these items.
                     if (scope.context.allDictionaryValues.length < result.totalCount) {
-                        $select.page++;
-
                         // Reset scrolling for the when-scrolled directive, so it could trigger this method for next page.
                         scope.$broadcast('scrollCompleted');
                     }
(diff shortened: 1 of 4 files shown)


VirtoCommerce.CatalogModule.Web/Scripts/directives/property2.js
```diff
@@ -116,28 +115,57 @@ angular.module('virtoCommerce.catalogModule')
...
-                    });
+                return scope.loadNextDictionaryValues($select);
+            };
+
+            scope.loadNextDictionaryValues = function($select) {
+                var countToSkip = $select.page * scope.pageSize;
+                var countToTake = scope.pageSize;
 
-                    //initLanguagesValuesMap();
+                return scope.getPropValues()(scope.currentEntity.id, '', countToSkip, countToTake).then(function (result) {
```
tatarincev: Search keyword doesn't take into accounts, it will be difficult to find appropriate dictionary value in the huge amount of data

avas: Sorry, I forgot about this. It is fixed now.

I also reduced `refresh-timeout` to 200 ms (default value was 1000 ms), so now search keyword will be applied quicker.

PR #186, opened 2018-12-13
va-property2 - hidden languages

Tilesto:
Added property for "invisible" languages


VirtoCommerce.CatalogModule.Web/Scripts/directives/property2.js
```diff
@@ -130,6 +131,13 @@ angular.module('virtoCommerce.catalogModule')
                 }
             };
 
+            scope.isLanguageVisible = function (language) {
```
tatarincev: I think you need to do additional null checks for  `scope.hiddenLanguages` here

Tilesto: Done

tatarincev (approved):

PR #192, opened 2019-01-29
fix #65 catalog issue

### LINKED ISSUE #65: The Catalog blade show wrong catalog & categories list when user is non-administrator
@asvishnyakov commented on [Wed Feb 22 2017](https://github.com/VirtoCommerce/vc-platform/issues/739)

How to reproduce:
1. Create new user
2. Assign role "Clothing accessory category manager" to it
3. Open catalog, try to browse catalogs. You will see pic 1. instead of pic. 2.
pic. 1
pic. 2
You may also try to assign "Electronic  store administrator" role.

---

@OlegoO commented on [Thu Oct 26 2017](https://github.com/VirtoCommerce/vc-platform/issues/739#issuecomment-339652258)

Issue moved to [VirtoCommerce/vc-module-catalog #64](https://github.com/VirtoCommerce/vc-module-catalog/issues/64) via [**ZenHub**](https://www.zenhub.com/)

Expected behavior:
- We should not see catalogs to which categories we don't have access for (ex. for "Clothing accessory category manager" only "Clothing" catalog should be visible);
- In case we have selected category scope, we should have permissions to its children ("Clothing accessory category manager" has catalog:read on "Accessories" category, means he is able to open child categories like "Bags" and products as well).
---
Technical details:
We could properly filter catalog SearchCriteria CatalogIds and CategoryIds according to user permission scopes in SearchCriteriaExtensions.ApplyRestrictionsForUser: modify SearchCriteria with the intersection of what we are trying to find and what we could see (note that CatalogSearchService returns both searched CatalogIds and CategoryIds content, not their intersection). 
In result some catalogs or categories may be filtered out if user has no access to them. E.g. if user search for A1 catalog and have scoped access only to C1 category inside A1 and to C2 category inside A2 catalog, we should add only C1 category to SearchCriteria CategoryIds, and exclude A1 catalog from SearchCriteria CatalogIds (as SearchService searches for catalogs and categories independently, and having A1 catalogId and C1 category in searchCriteria would lead to returning content of A1 catalog **AND** C1 content, when we have access **ONLY** to C1 category content).

Also need to add unit tests for the filtering method.

anivirtoway:
Fix the Catalog blade show wrong catalog & categories list when user is non-administrator #65
catalog


yecli:
Created under wrong account, please recreate

PR #200, opened 2019-02-22
Force parameterized offset query to fix paging issue.

merlevedeb:
When searching for all property dictionary items using the propertyDictionaryItemService, it seems that a combination of sorting and paging on the produced EF query returns unstable (duplicate) results.

This behaviour got introduced when an additional sort was added in commit https://github.com/VirtoCommerce/vc-module-catalog/commit/56fb799d

I was able to mitigate the issue by forcing a parameterized query but the root cause why SQL is returning unstable results is unknown to me.

Reproducible by executing the queries in following file and comparing the result sets:
[EFReproductionQueries.zip](https://github.com/delawarePro/vc-module-catalog/files/2892976/EFReproductionQueries.zip)

### DIFF
--- a/VirtoCommerce.CatalogModule.Data/Services/PropertyDictionaryItemService.cs
+++ b/VirtoCommerce.CatalogModule.Data/Services/PropertyDictionaryItemService.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Data.Entity;
 using System.Linq;
 using CacheManager.Core;
 using VirtoCommerce.CatalogModule.Data.Model;
@@ -130,7 +131,7 @@ namespace VirtoCommerce.CatalogModule.Data.Services
                     query = query.OrderBySortInfos(sortInfos);
 
                     result.TotalCount = query.Count();
-                    var ids = query.Skip(criteria.Skip).Take(criteria.Take).Select(x => x.Id).ToArray();
+                    var ids = query.Skip(() => criteria.Skip).Take(() => criteria.Take).Select(x => x.Id).ToArray();
                     result.Results = GetByIds(ids).AsQueryable().OrderBySortInfos(sortInfos).ToList();
                     return result;
                 }


yecli:
Difference in query generation is described here (EF needs to build every time for each parameter value, but only once for expression taking parameter):
https://visualstudiomagazine.com/articles/2016/12/06/skip-take-entity-framework-lambda.aspx

Thanks for useful optimization idea!

yecli:
What for duplicate rows - checked your example queries, it seems that we sort based on Alias and SortOrder columns. And there are multiple rows with this values duplicated, like:

> Line 65: INSERT INTO #PropertyDictionaryItem ([Id], [Alias], [PropertyId], [SortOrder]) VALUES (N'064d469d21d64b9e95cc6b065cec74e9', N'Yes', N'E7743053-FD3C-47AC-80FE-A8BC1A0EA775', 0)
> 	Line 115: INSERT INTO #PropertyDictionaryItem ([Id], [Alias], [PropertyId], [SortOrder]) VALUES (N'228ce5b2f0de4761a375577810268c25', N'Yes', N'B828CABF-A4BE-4772-A020-A464C1C3D3DE', 0)
> 	Line 125: INSERT INTO #PropertyDictionaryItem ([Id], [Alias], [PropertyId], [SortOrder]) VALUES (N'27ba6319f6d54c199c6a5f4aae0fac33', N'Yes', N'E890D13D-10C2-48AD-955A-5E33EF06B4A1', 0)


	and so on

In such case ordering is non deterministic ([proof](https://stackoverflow.com/questions/35047119/order-by-boolean-column-produce-duplicate-rows-when-paging-in-sql-server)).

Another question - why your changes could help avoid duplicates? 

In general, i think it is still possible to get duplicates even with the fix, but the probability is reduced.
I could assume that in case of multiple different queries that are compiled every time we could get different execution plans, which could lead to different order. Whereas for once compiled query with params we got 1 execution plan (or at least chances that it would be changed are lesser), so it is more stable. Maybe even 1 execution plan could provide different results in case of non-deterministic sorting.

To fully eliminate the possibility of duplicate values we need to use deterministic ordering (like add ThenBy(Id)).

We will approve your changes and include in upcoming release, and will think about this problem.

merlevedeb:
Our aliases are only unique per property so indeed the better change is to add the id in the sorting as well.
The behavior makes sense now and indeed it could still happen with the code change. 
Thanks for the elaborate explanation and considering the deterministic order.

yecli:
Released in [v2.23.14](https://github.com/VirtoCommerce/vc-module-catalog/releases/tag/v2.23.14)

PR #207, opened 2019-03-20
fix Invalid object name 'SeoUrlKeyword'

AliveMen:
move deleting SeoUrlKeyword from SQL to service

### DIFF
--- a/VirtoCommerce.CatalogModule.Data/Services/CategoryServiceImpl.cs
+++ b/VirtoCommerce.CatalogModule.Data/Services/CategoryServiceImpl.cs
@@ -100,6 +100,11 @@ namespace VirtoCommerce.CatalogModule.Data.Services
                 .Select(c => new GenericChangedEntry<Category>(c, EntryState.Deleted))
                 .ToList();
 
+            foreach (var category in categories)
+            {
+                _commerceService.DeleteSeoForObject(category);
+            }
+
             using (var repository = _repositoryFactory())
             {
                 _eventPublisher.Publish(new CategoryChangingEvent(changedEntries));
--- a/VirtoCommerce.CatalogModule.Data/Services/CategoryServiceImpl.cs
+++ b/VirtoCommerce.CatalogModule.Data/Services/CategoryServiceImpl.cs
@@ -100,7 +100,10 @@ namespace VirtoCommerce.CatalogModule.Data.Services
                 .Select(c => new GenericChangedEntry<Category>(c, EntryState.Deleted))
                 .ToList();
 
-            _commerceService.DeleteSeoForObjects(categories);
+            foreach (var category in categories)
+            {
+                _commerceService.DeleteSeoForObject(category);
+            }
 
             using (var repository = _repositoryFactory())
             {
--- a/VirtoCommerce.CatalogModule.Data/Services/CategoryServiceImpl.cs
+++ b/VirtoCommerce.CatalogModule.Data/Services/CategoryServiceImpl.cs
@@ -100,10 +100,7 @@ namespace VirtoCommerce.CatalogModule.Data.Services
                 .Select(c => new GenericChangedEntry<Category>(c, EntryState.Deleted))
                 .ToList();
 
-            foreach (var category in categories)
-            {
-                _commerceService.DeleteSeoForObject(category);
-            }
+            _commerceService.DeleteSeoForObjects(categories);
 
             using (var repository = _repositoryFactory())
             {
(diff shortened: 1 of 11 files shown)


VirtoCommerce.CatalogModule.Data/Services/CategoryServiceImpl.cs
```diff
@@ -100,6 +100,11 @@ public virtual void Delete(string[] categoryIds)
                 .Select(c => new GenericChangedEntry<Category>(c, EntryState.Deleted))
                 .ToList();
 
+            foreach (var category in categories)
```
yecli: Here i think we need to add method DeleteSeoForObject**s** for array of objects to ICommerceService, and make current DeleteSeoForObject obsolete. It would allow us to avoid multiple requests

VirtoCommerce.CatalogModule.Data/Services/CategoryServiceImpl.cs
```diff
@@ -100,6 +100,8 @@ public virtual void Delete(string[] categoryIds)
                 .Select(c => new GenericChangedEntry<Category>(c, EntryState.Deleted))
                 .ToList();
 
+            _commerceService.DeleteSeoForObjects(categories);
```
yecli: Decided to not add new method to this interface to avoid possible compatibility problems with existing custom implementations, so please use existing `DeleteSeoForObject`.

VirtoCommerce.CatalogModule.Data/Services/ItemServiceImpl.cs
```diff
@@ -116,6 +116,8 @@ public virtual void Delete(string[] itemIds)
                 .Select(i => new GenericChangedEntry<CatalogProduct>(i, EntryState.Deleted))
                 .ToList();
 
+            _commerceService.DeleteSeoForObjects(items);
```
yecli: Decided to not add new method to this interface to avoid possible compatibility problems with existing custom implementations, so please use existing `DeleteSeoForObject`.

VirtoCommerce.CatalogModule.Data/Services/ItemServiceImpl.cs
```diff
@@ -116,6 +116,10 @@ public virtual void Delete(string[] itemIds)
                 .Select(i => new GenericChangedEntry<CatalogProduct>(i, EntryState.Deleted))
                 .ToList();
 
+            foreach (var catalogProduct in items)
```
tatarincev: I'll suggest to move this logic after the products deletion

VirtoCommerce.CatalogModule.Test/packages.config
```diff
@@ -14,7 +14,7 @@
   <package id="System.ComponentModel.Annotations" version="4.4.1" targetFramework="net461" />
   <package id="System.ComponentModel.Primitives" version="4.3.0" targetFramework="net461" />
   <package id="valueinjecter" version="2.3.3" targetFramework="net451" />
-  <package id="VirtoCommerce.CoreModule.Data" version="2.25.16" targetFramework="net461" />
+  <package id="VirtoCommerce.CoreModule.Data" version="2.25.18" targetFramework="net461" />
```
tatarincev: Why do we need to update the CoreModule.Data version for this task?

VirtoCommerce.CatalogModule.Data/Services/CategoryServiceImpl.cs
```diff
@@ -111,6 +111,11 @@ public virtual void Delete(string[] categoryIds)
 
                 _eventPublisher.Publish(new CategoryChangedEvent(changedEntries));
             }
+
+            foreach (var category in categories)
```
tatarincev: This logic doesn't remove the seo info  for nested categories need to expand categories tree and call DeleteSeoForObject for these objects.

VirtoCommerce.CatalogModule.Data/Services/CategoryServiceImpl.cs
```diff
@@ -100,19 +100,30 @@ public virtual void Delete(string[] categoryIds)
                 .Select(c => new GenericChangedEntry<Category>(c, EntryState.Deleted))
                 .ToList();
 
+            Category[] children;
```
artem-dudarev: Separate with empty line

yecli: Done

artem-dudarev:
Catalog module should not know anything about SEO table

PR #208, opened 2019-03-27
Paged search ordering

AliveMen:
add thenById


VirtoCommerce.CatalogModule.Data/Services/CatalogSearchServiceImpl.cs
```diff
@@ -133,7 +131,7 @@ protected virtual void SearchCategories(SearchCriteria criteria, SearchResult re
                 //Try to replace sorting columns names
                 TryTransformSortingInfoColumnNames(_categorySortingAliases, sortInfos);
 
-                query = query.OrderBySortInfos(sortInfos);
+                query = query.OrderBySortInfos(sortInfos).ThenBy(q => q.Id);
```
yecli: According to naming style need to use x for such linq with one argument (in other cases too).

VirtoCommerce.CatalogModule.Data/Extensions/SortInfoExtension.cs
```diff
@@ -0,0 +1,25 @@
...
+
+namespace VirtoCommerce.CatalogModule.Data.Extensions
+{
+    public static class SortInfoExtension
+    {
+        public static SortInfo[] EnsureIdIncluded(this SortInfo[] sortInfos)
+        {
+            if (!sortInfos.IsNullOrEmpty())
+            {
+                if (!sortInfos.Any(x => x.SortColumn.EqualsInvariant("id")))
```
artem-dudarev: Move to the previous if()

VirtoCommerce.CatalogModule.Data/Extensions/SortInfoExtension.cs
```diff
@@ -0,0 +1,25 @@
...
+                {
+                    var sortInfoLst = sortInfos.ToList();
```
artem-dudarev: There is no such word as 'Lst'

VirtoCommerce.CatalogModule.Data/Extensions/SortInfoExtension.cs
```diff
@@ -0,0 +1,25 @@
...
+                    sortInfoLst.Add(new SortInfo()
+                    {
+                        SortColumn = "id"
+                    });
+                    return sortInfoLst.ToArray();
```
artem-dudarev: Single return

VirtoCommerce.CatalogModule.Data/Extensions/SortInfoExtension.cs
```diff
@@ -7,19 +8,16 @@ public static class SortInfoExtension
     {
         public static SortInfo[] EnsureIdIncluded(this SortInfo[] sortInfos)
         {
-            if (!sortInfos.IsNullOrEmpty())
+            var sortInfoList = sortInfos?.ToList() ?? new List<SortInfo>();
```
artem-dudarev: Don't create new list if original sortInfos array already contains 'id'

AliveMen: I'm create new list when original sortInfos == null only

artem-dudarev: `ToList()` creates new list

yecli: Probably Artem meant we could change the result under condition
`if (!sortInfoList?.Any(x => x.SortColumn.EqualsInvariant("id")))`

artem-dudarev: I mean there is no need to create new list and then convert it back to array if we already have 'id' in original array.

VirtoCommerce.CatalogModule.Data/Extensions/SortInfoExtension.cs
```diff
@@ -0,0 +1,35 @@
+using System.Collections.Generic;
+using System.Linq;
+using VirtoCommerce.Platform.Core.Common;
+
+namespace VirtoCommerce.CatalogModule.Data.Extensions
+{
+    public static class SortInfoExtension
+    {
+        public static SortInfo[] EnsureIdIncluded(this SortInfo[] sortInfos)
```
artem-dudarev:
```csharp
public static SortInfo[] EnsureIdIncluded(this SortInfo[] sortInfos)
{
    var result = sortInfos;

    if (sortInfos == null || !sortInfos.Any(x => x.SortColumn.EqualsInvariant("id")))
    {
        var list = sortInfos?.ToList() ?? new List<SortInfo>();
        list.Add(new SortInfo { SortColumn = "id" });
        result = list.ToArray();
    }

    return result;
}
```

yecli:
@tatarincev I think we could use that extensions to platform and use this `OrderBySortInfosWithId` method as generic solution.

Issue #204, opened 2019-03-12
Variation dictionary properties are not restored correctly after import/export


yecli:
Color property of Cell phones is not saved after platform export/import procedure.

### Steps to reproduce
1. Set Color property for phone variation, Save variation:
![image](https://user-images.githubusercontent.com/15610352/54216932-71a3e000-450c-11e9-9b46-6fa300f331cf.png)
2. Ensure it is visible in storefront Cell phone filters
![image](https://user-images.githubusercontent.com/15610352/54217410-63a28f00-450d-11e9-9c03-cbe2b05853c5.png)
3. Export Catalog binary data
![image](https://user-images.githubusercontent.com/15610352/54216742-0eb24900-450c-11e9-9d00-b7b9fcea8967.png)
4. Import into same database (or new one).
5. Check the property in storefront filter - there is no Sheer gold
![image](https://user-images.githubusercontent.com/15610352/54217377-52598280-450d-11e9-9dba-de20195ba666.png)

###Technical details:
Seems that empty ShortTextName for Variation Color PropertyValue leads to 'undef' in storefront filter. Seems empty Alias leads to this ShortTextValue.

PropertyValue table before export:
![image](https://user-images.githubusercontent.com/15610352/54217262-1aead600-450d-11e9-94e9-a95469d2cf62.png)
PropertyValue table before import:
![image](https://user-images.githubusercontent.com/15610352/54217289-2b9b4c00-450d-11e9-93d8-4afe4aac7c3e.png)

Property value for main product on export:
![image](https://user-images.githubusercontent.com/15610352/54217465-7ddc6d00-450d-11e9-9bde-3cde3005e379.png)
Property value for variation (no Alias):
![image](https://user-images.githubusercontent.com/15610352/54217515-95b3f100-450d-11e9-8a00-7a8b8c2333e5.png)

Catalog Json:
Main product:
![image](https://user-images.githubusercontent.com/15610352/54217548-acf2de80-450d-11e9-9715-36208921816d.png)
Variation - no Alias:
![image](https://user-images.githubusercontent.com/15610352/54217610-c4ca6280-450d-11e9-834a-ae4062ae667e.png)

yecli:
Related to https://github.com/virtocommerce/vc-theme-default/issues/58

kostyrin:
[...]

PR #212, opened 2019-04-01
#204 Variation dictionary properties are not restored correctly after import/export:

yecli:
- Load property dictionary items and values based on ResponseGroup;


yecli:
Decided to load properties with dictionary items always, without ResponseGroup check.

PR #216, opened 2019-04-11
Add bool IsHidden property to catalog property

AliveMen:
AAI_190411_125303
add isHiddenProperty
add migration
add field to UI
update Core dependency


VirtoCommerce.CatalogModule.Web.Core/Model/Property.cs
```diff
@@ -32,6 +32,7 @@ public Property(PropertyValue propValue, string catalogId, PropertyType property
             Type = propertyType;
             ValueType = propValue.ValueType;
             Values = new List<PropertyValue>();
+            Hidden = false;
```
artem-dudarev: This can be removed. Default value is `false`.

PR #219, opened 2019-04-17
added field MainProductId to WebModel product

adminoren:


tatarincev:
@adminoren The current product web model already has TitualProductId property with the same meaning as MainProductId in domain product model, please use it on the storefront side.
https://github.com/VirtoCommerce/vc-module-catalog/blob/dev/VirtoCommerce.CatalogModule.Web.Core/Converters/ProductConverter.cs#L117

Issue #221, opened 2019-04-19
Product indexed search gives unexpected order when we are using sorting by non-existent index attribute


yecli:
### Actual behavior
If we use default Featured sorting in storefront, we have "manual" attribute for sorting. It is not presented in index attributes but is passed to SearchProvider. It leads to unclear sort results on Storefront.

### Expected behavior
Probably in that case we should use some default sorting, e.g. by [priority, __sort == "name" now].

**Technical details:**
Probably [this line](https://github.com/VirtoCommerce/vc-module-catalog/blame/dev/VirtoCommerce.CatalogModule.Data/Search/ProductSearchRequestBuilder.cs#L89) show be removed, and this code should be executed:
```
            if (!result.Any())
            {
                result.AddRange(priorityFields.Select(priorityField => new SortingField(priorityField, true)));
                result.Add(new SortingField("__sort"));
            }
```

yecli:
This problem should be fixed in storefront.

PR #223, opened 2019-04-22
#221 Product indexed search gives unexpected order when we are using sorting by non-existent index attribute:

yecli:
- Added "manual" field handling (corresponds to "Featured" on storefront);


yecli:
After discussion with @artem-dudarev, decided to explicitly pass proper sort fields in storefront instead of using "hardcoded" values on platform level.

PR #225, opened 2019-04-24
fix Invalid object name 'SeoUrlKeyword'

AliveMen:
check if SeoUrlKeyword is exist

### DIFF
--- a/VirtoCommerce.CatalogModule.Data/Repositories/CatalogRepositoryImpl.cs
+++ b/VirtoCommerce.CatalogModule.Data/Repositories/CatalogRepositoryImpl.cs
@@ -400,8 +400,9 @@ namespace VirtoCommerce.CatalogModule.Data.Repositories
             if (!itemIds.IsNullOrEmpty())
             {
                 const string commandTemplate = @"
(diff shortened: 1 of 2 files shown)


VirtoCommerce.CatalogModule.Data/Repositories/CatalogRepositoryImpl.cs
```diff
@@ -399,9 +399,10 @@ public void RemoveItems(string[] itemIds)
         {
             if (!itemIds.IsNullOrEmpty())
             {
-                const string commandTemplate = @"
+
+                var commandTemplate = (IsSeoUrlKeywordExists() ? @"
                     DELETE SEO FROM SeoUrlKeyword SEO INNER JOIN Item I ON I.Id = SEO.ObjectId AND SEO.ObjectType = 'CatalogProduct'
```
artem-dudarev:
```SQL
IF EXISTS (SELECT 1 FROM sys.tables WHERE [Name] = 'SeoUrlKeyword')
    DELETE ...
```

AliveMen: fixed

yecli:
Agree with you. As I understand, SeoUrlKeyword logic will be totally reworked in platform v3. By this change we just got rid of this error.

PR #257, opened 2019-09-13
fix vp-511 Problem with title of screen

AliveMen:
add missed titles

### DIFF
--- a/VirtoCommerce.CatalogModule.Web/Localizations/en.VirtoCommerce.Catalog.json
+++ b/VirtoCommerce.CatalogModule.Web/Localizations/en.VirtoCommerce.Catalog.json
@@ -466,8 +466,8 @@
         }
       },
       "exporter-list": {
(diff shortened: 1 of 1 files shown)


VirtoCommerce.CatalogModule.Web/Localizations/en.VirtoCommerce.Catalog.json
```diff
@@ -464,7 +464,15 @@
           "placeholder": "Add number",
           "description": "Bounds will be sorted automatically before saving. Price ranges are constructed as follows: Under [bound 1], [bound 1] to [bound 2], ... , Over [bound N]"
         }
-      }
+      },
+      "exporter-list": {
+        "title": "Available export modules",
```
yecli: I think for consistency we should use Provider
```
        "title": "Available export providers",
        "subtitle": "Select provider to export products"
```

PR #267, opened 2019-09-24
vp_525_Remove_inconsistencies_in_catalog_export

AliveMen:
use isAllSelected and isAnyFilterApplied


yecli:
Depends on VirtoCommerce/vc-module-export#22 for proper filter passing;

PR #270, opened 2019-09-25
dev_486_traffic_optimization

AliveMen:
update dependency version
use flag isPreview


VirtoCommerce.CatalogModule.Data/ExportImport/ProductFullExportDataQuery.cs
```diff
@@ -35,7 +36,7 @@ public virtual ProductExportDataQuery ToProductExportDataQuery()
             result.Skip = Skip;
             result.Sort = Sort;
             result.Take = Take;
-            result.LoadImageBinaries = true;
+            result.IsPreview = IsPreview;
```
yecli: Do we need this one if we fill it in vase class?

PR #278, opened 2019-10-16
#vp732 IndexCategoryChangedEventHandler IndexProductChangedEventHandler

kostyrin:


src/VirtoCommerce.CatalogModule.Data/Handlers/IndexCategoryChangedEventHandler.cs
```diff
@@ -0,0 +1,54 @@
...
+        }
+
+        public Task Handle(CategoryChangedEvent message)
+        {
+            if (message == null)
+            {
+                throw new ArgumentNullException(nameof(message));
+            }
+
+            var indexCategoryIds = message.ChangedEntries.Where(x => (x.EntryState == EntryState.Modified || x.EntryState == EntryState.Added) && x.OldEntry.Id != null)
```
tatarincev: Why we van't handle deleted product as well as updated?

kostyrin: we can

PR #289, opened 2019-10-29
Bulk update/move models to up level

t13ka:
The changes in order to release the bulk actions:
- ListEntries models were moved to the web.core project
- old logic which was in the controller moved into dedicated services
- no unit tests yet

similar https://github.com/VirtoCommerce/vc-module-catalog/tree/bulk-update/dev/VirtoCommerce.CatalogModule.Web.Core/Model


VirtoCommerce.CatalogModule.Data/Services/CategoryMover.cs
```diff
@@ -0,0 +1,57 @@
+using System.Collections.Generic;
+using System.Linq;
+
+using VirtoCommerce.CatalogModule.Web.Model;
+using VirtoCommerce.CatalogModule.Web.Services;
+using VirtoCommerce.Domain.Catalog.Services;
+using VirtoCommerce.Platform.Core.Common;
+
+using VC = VirtoCommerce.Domain.Catalog.Model;
```
yecli: Not clear what means VC. "domain" or "coreModel" would be self-explanatory, and will be coherent to the existing naming style.

VirtoCommerce.CatalogModule.Data/Services/CategoryMover.cs
```diff
@@ -0,0 +1,57 @@
...
+        public override void ConfirmMove(IEnumerable<VC.Category> entities)
+        {
+            _categoryService.Update(entities.ToArray());
+        }
+
+        public override List<VC.Category> PrepareMove(MoveContext moveContext)
+        {
+            var result = new List<VC.Category>();
+
+            foreach (var listEntryCategory in moveContext.ListEntries.Where(
```
yecli: No need to break the line, it is just one pretty short LINQ statement (less than 200 symbols).

VirtoCommerce.CatalogModule.Data/Services/ListEntrySearchService.cs
```diff
@@ -0,0 +1,91 @@
...
+                result.TotalCount += searchResult.ProductsTotalCount;
+
+                result.ListEntries.AddRange(products);
+            }
+
+            return result;
+        }
+
+        private static IEnumerable<T> GetPaged<T>(IEnumerable<T> collection, int skip, int take)
+        {
```
yecli: For me having that method reduces readability (as I do not understand what it does by its name and I need to check the implementation). If we have only one usage of this code, I do not see the reason to decrease readability if we are not avoiding duplication.

VirtoCommerce.CatalogModule.Data/Services/ProductMover.cs
```diff
@@ -0,0 +1,75 @@
+using System.Collections.Generic;
+using System.Linq;
+
+using VirtoCommerce.CatalogModule.Web.Model;
+using VirtoCommerce.CatalogModule.Web.Services;
+using VirtoCommerce.Domain.Catalog.Services;
+using VirtoCommerce.Platform.Core.Common;
+using VC = VirtoCommerce.Domain.Catalog.Model;
```
yecli: Unclear naming for the namespace.

VirtoCommerce.CatalogModule.Web.Core/Model/MoveContext.cs
```diff
@@ -1,18 +1,13 @@
...
-using System.Linq;
-using System.Web;
 
 namespace VirtoCommerce.CatalogModule.Web.Model
 {
-    /// <summary>
-    /// Represent move operation detail
-    /// </summary>
-    public class MoveInfo
+    public class MoveContext
```
yecli: I think we should not rename existing classes, as it brakes backward compatibility. And in general I do not see the real difference between MoveInfo and MoveContext.
Could you just rename the file to MoveInfo?

Issue #279, opened 2019-10-17
Ability to use image URLs as product images directly without uploading it to the storage


yecli:
Need to have an ability to use image URLs as product images directly without uploading it to the storage.
![image](https://user-images.githubusercontent.com/15610352/66987002-52d87e00-f0d9-11e9-843b-f3eceb62b35a.png)

We could add "fa-external-link" icon before "+" with title "Use image from URL", that will save asset URL without uploading it.

tatarincev:
Merged and released in https://github.com/VirtoCommerce/vc-module-catalog/releases/tag/v2.23.29 but due to that some UX improvement still pending we postpone the including this functionality in the release notes

yecli:
Released in https://github.com/VirtoCommerce/vc-module-catalog/releases/tag/v2.23.31

PR #291, opened 2019-11-15
add ability to use image url directly without upload

ilyawzrd:
for https://github.com/VirtoCommerce/vc-module-catalog/issues/279
uploadUrl


VirtoCommerce.CatalogModule.Web/Scripts/blades/images-add.tpl.html
```diff
@@ -52,7 +52,10 @@
                 </div>
 
                 <div class="form-input __file">
-                    <input ng-model="blade.newExternalImageUrl" type="url" placeholder="{{ 'catalog.blades.images.placeholders.image-url' | translate }}" tabindex="-1">
+                    <input ng-model="blade.newExternalImageUrl" type="url" placeholder="{{ 'catalog.blades.images.placeholders.image-url' | translate }}" tabindex="-1" style="padding-right: 80px">
```
tatarincev: The not strict comment, but in the future need to avoid of inline styles usage and use css classes instead

PR #296, opened 2019-11-22
Added checking for null before working with property values

pushnitsa:
Issue comes when we try to create item in the IItemService.Create() with object which have PropertyValues = null.
Dear colleagues, could you please tell me, why we do not initializing collections in the constructors?


yecli:
>>Dear colleagues, could you please tell me, why we do not initializing collections in the constructors?

If you meant initializing this collection `Product.PropertyValues` in the model - probably because we want this null value to be the marker that collection should not be changed in entity.

yecli:
Released in https://github.com/VirtoCommerce/vc-module-catalog/releases/tag/v2.23.31

Issue #300, opened 2019-12-09
Move logic away from ApiCatalogBrowseFiltersController to make it extendable


asvishnyakov:
https://github.com/VirtoCommerce/vc-module-catalog/blob/f0f0cd5751a6e5641134e40ad2aa69dad17cd8cf/VirtoCommerce.CatalogModule.Web/Controllers/Api/CatalogBrowseFiltersController.cs#L128-L137

### Expected behavior
We should have ability to extend the list of filtration properties

### Actual behavior
We can't do that.

### Steps to reproduce
-

Woland2k:
Would be helpful to understand the scenario for which it needs to be extended.

asvishnyakov:
@Woland2k We want to add custom aggregation property to receive it as part of aggregations of product in storefront and show the number of products available in filter, for example. Right now the only thing we can do is to hard code it in theme or "emulate" aggregations in storefront

PR #302, opened 2019-12-16
#300 GetAllProperties move from CatalogBrowseFiltersController

kostyrin:


t13ka:
added to the internal backlog

PR #324, opened 2020-02-05
fix: propertydictionaryitem api response type

yongbin-kooboo:
The PropertyDictionaryItem search api response type is 'PropertyDictionaryItem[]' not 'PropertyDictionaryItemSearchResult', it causes that autorest client api deserialize json failed.


zhangpengchen:
Hi @tatarincev, 

Is it something you could help to review, we are blocking with the error at this moment.

PR #326, opened 2020-02-17
vp-1419 migrate CatalogPersonalization module to V3

AliveMen:
Use own search request builder registrar for services registration instead Core IoC.
Default IoC can't override already registered service by new one


AliveMen:
Move searchRequestBuilderRegistrar to the Search module.

PR #352, opened 2020-04-29
VP-2065 Fix price blade error when in platform registered more than 20 catalogs

AliveMen:
Change search criteria to get all catalogues and reduce response group.

### DIFF
--- a/src/VirtoCommerce.CatalogModule.Web/Controllers/Api/CatalogModuleCatalogsController.cs
+++ b/src/VirtoCommerce.CatalogModule.Web/Controllers/Api/CatalogModuleCatalogsController.cs
@@ -41,13 +41,12 @@ namespace VirtoCommerce.CatalogModule.Web.Controllers.Api
         [HttpGet]
         [Route("")]
         [Obsolete("use POST api/catalog/catalogs/search instead")]
-        public async Task<ActionResult<Catalog[]>> GetCatalogs(string sort = null, int skip = 0, int take = 1000)
+        public async Task<ActionResult<Catalog[]>> GetCatalogs(string sort = null, int skip = 0, int take = 20)
         {
             var criteria = AbstractTypeFactory<CatalogSearchCriteria>.TryCreateInstance();
             criteria.Sort = sort;
             criteria.Skip = skip;
             criteria.Take = take;
-            criteria.ResponseGroup = CatalogResponseGroup.Info.ToString();
 
             var authorizationResult = await _authorizationService.AuthorizeAsync(User, criteria, new CatalogAuthorizationRequirement(ModuleConstants.Security.Permissions.Read));
             if (!authorizationResult.Succeeded)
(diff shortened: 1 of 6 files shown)


src/VirtoCommerce.CatalogModule.Web/Controllers/Api/CatalogModuleCatalogsController.cs
```diff
@@ -39,12 +39,13 @@ ICatalogService catalogService
         /// <remarks>Get common and virtual Catalogs list with minimal information included. Returns array of Catalog</remarks>
 		[HttpGet]
         [Route("")]
-        public async Task<ActionResult<Catalog[]>> GetCatalogs(string sort = null, int skip = 0, int take = 20)
+        public async Task<ActionResult<Catalog[]>> GetCatalogs(string sort = null, int skip = 0, int take = 1000)
```
yecli, line 42: Shouldn't we change that condition on the client side?
I think it is safer to explicitly specify this value on the client if we need more.

src/VirtoCommerce.CatalogModule.Web/Controllers/Api/CatalogModuleCatalogsController.cs
```diff
@@ -39,12 +39,13 @@ ICatalogService catalogService
...
         {
             var criteria = AbstractTypeFactory<CatalogSearchCriteria>.TryCreateInstance();
             criteria.Sort = sort;
             criteria.Skip = skip;
             criteria.Take = take;
+            criteria.ResponseGroup = CatalogResponseGroup.Info.ToString();
```
yecli, line 48: @tatarincev 
Should we set minimal response group here according to the method description - "Get common and virtual Catalogs list with minimal information included"(as it is Full by default)?

src/VirtoCommerce.CatalogModule.Web/Controllers/Api/CatalogModuleCatalogsController.cs
AliveMen: reverted

src/VirtoCommerce.CatalogModule.Web/Controllers/Api/CatalogModuleCatalogsController.cs
AliveMen: reverted

src/VirtoCommerce.CatalogModule.Web/Scripts/widgets/dashboard.js
```diff
@@ -1,13 +1,13 @@
...
 .controller('virtoCommerce.catalogModule.dashboard.catalogsWidgetController', ['$scope', '$state', 'virtoCommerce.catalogModule.catalogs', function ($scope, $state, catalogs) {
     $scope.data = { count: '', descr: 'Catalogs' };
 
     $scope.widgetAction = function () {
         $state.go('workspace.catalog');
     };
 
-    catalogs.getCatalogs({}, function (data) {
-        $scope.data.count = data.length;
+    catalogs.search({take:1000}, function (data) {
```
yecli, line 9: Here I think we should specify take:0, as we only using length. Please check, possible redundant queries

AliveMen: fixed

PR #354, opened 2020-05-06
VP-2145 Fix permission error when displaying properties

AliveMen:
Add authorization handler for PropertyDictionaryItemSearchCriteria


src/VirtoCommerce.CatalogModule.Web/Controllers/Api/CatalogModulePropertyDictionaryItemsController.cs
```diff
@@ -30,9 +34,13 @@ public class CatalogModulePropertyDictionaryItemsController : Controller
         /// <returns></returns>
         [HttpPost]
         [Route("search")]
-        [Authorize(ModuleConstants.Security.Permissions.Read)]
```
tatarincev, line 33: Please leave empty [Authorize] attribute for this method. It is required to swagger ui.

AliveMen: done

PR #373, opened 2020-05-28
Catalog performance  optimization

tatarincev:


lnetrebskii:
Review task https://virtocommerce.atlassian.net/browse/VP-2637 has been created

src/VirtoCommerce.CatalogModule.Core/Extensions/EnumerableExtensions.cs
```diff
@@ -0,0 +1,29 @@
+using System;
```
yecli, line 1: The same method was added into the platform. Maybe need to remove it from here.

akak1977: Yes we can. But we need to raise the platform dependency to an unreleased version.

yecli: Or release the platform first, and then use it here.

PR #382, opened 2020-06-09
VP-2761 Fix IndexChangesJob does not index new Imported productds.

trueboroda:
When the event base indexation is turned off.


src/VirtoCommerce.CatalogModule.Data/Search/Indexing/ProductDocumentChangesProvider.cs
```diff
@@ -39,15 +39,24 @@ public virtual async Task<long> GetTotalChangesCountAsync(DateTime? startDate, D
             }
             else
             {
+                // Get added and modified products count
+                using (var repository = _catalogRepositoryFactory())
+                {
+                    result = await repository.Items.CountAsync(i => i.ParentId == null && ((i.ModifiedDate >= startDate && i.ModifiedDate <= endDate) || (i.CreatedDate >= startDate && i.CreatedDate <= endDate)));
```
yecli, line 45: IS this query working as expected if startDate or endDate = null?

trueboroda: It is an excepted case. In that case, the system should throw an argument exception.  It was done.

yecli: The same validation should be done in GetTotalChangesCountAsync, isn't it?

src/VirtoCommerce.CatalogModule.Data/Search/Indexing/ProductDocumentChangesProvider.cs
```diff
@@ -56,56 +52,112 @@ public virtual async Task<long> GetTotalChangesCountAsync(DateTime? startDate, D
...
+
+                    var modifiedProductIndexDocumentChanges = await GetModifiedProductIndexDocumentChanges(startDate, endDate, skip, take, repository);
+                    result.AddRange(modifiedProductIndexDocumentChanges);
+                }
             }
 
             return result;
         }
+
+        private async Task<ChangeLogSearchResult> SearchDeleteOperationIntoLog(DateTime? startDate, DateTime? endDate, long skip, long take)
```
tatarincev, line 93: SearchDeleteOperationIntoLog -> SearchDeleteOperationInLog

PR #397, opened 2020-06-19
VP-2101: Dynamic product associations

yecli:
Dynamic product associations functionality


Woland2k:
shouldn't this be a separate module instead of being part of the catalog?

yecli:
We will move this functionality to the separate module: vc-module-dynamic-associations

PR #409, opened 2020-07-23
Fix MS SQL Error - The incoming request has too many parameters. 

OlegoO:
Fix MS SQL Error - The incoming request has too many parameters. server supports a maximum of 2100 parameters if remove more than 2100 products.


mvktsk:
Review task https://virtocommerce.atlassian.net/browse/VP-3778 has been created

yecli:
Agree, probably we need to make deletion in a transaction.
Added @tatarincev to review

OlegoO:
Let create delete in transactions as separated epic

PR #421, opened 2020-09-08
VP-2605: Fix 404 when delete all selected products

AliveMen:
### Problem
Error when try to delete more then 50 products 

### Solution
Add API for delete plenty catalog products

### DIFF
--- a/src/VirtoCommerce.CatalogModule.Web/Controllers/Api/CatalogModuleListEntryController.cs
+++ b/src/VirtoCommerce.CatalogModule.Web/Controllers/Api/CatalogModuleListEntryController.cs
@@ -16,7 +16,6 @@ using VirtoCommerce.CatalogModule.Data.Services;
 using VirtoCommerce.CatalogModule.Web.Model;
 using VirtoCommerce.Platform.Core.Common;
 using VirtoCommerce.Platform.Core.Settings;
-using VirtoCommerce.SearchModule.Core.Model;
 
 namespace VirtoCommerce.CatalogModule.Web.Controllers.Api
 {
@@ -259,18 +258,15 @@ namespace VirtoCommerce.CatalogModule.Web.Controllers.Api
                     .ToList();
             }
 
-            if (!idsToDelete.IsNullOrEmpty())
+            for (var i = 0; i < idsToDelete.Count; i += deleteBatchSize)
             {
-                for (var i = 0; i < idsToDelete.Count; i += deleteBatchSize)
-                {
-                    var commonIds = idsToDelete.Skip(i).Take(deleteBatchSize).ToArray();
+                var commonIds = idsToDelete.Skip(i).Take(deleteBatchSize).ToArray();
 
-                    var searchProductResult = await _itemService.GetByIdsAsync(commonIds, ItemResponseGroup.None.ToString());
-                    await _itemService.DeleteAsync(searchProductResult.Select(x => x.Id).ToArray());
+                var searchProductResult = await _itemService.GetByIdsAsync(commonIds, ItemResponseGroup.None.ToString());
+                await _itemService.DeleteAsync(searchProductResult.Select(x => x.Id).ToArray());
 
-                    var searchCategoryResult = await _categoryService.GetByIdsAsync(commonIds, CategoryResponseGroup.None.ToString());
-                    await _categoryService.DeleteAsync(searchCategoryResult.Select(x => x.Id).ToArray());
-                }
+                var searchCategoryResult = await _categoryService.GetByIdsAsync(commonIds, CategoryResponseGroup.None.ToString());
+                await _categoryService.DeleteAsync(searchCategoryResult.Select(x => x.Id).ToArray());
             }
 
             return StatusCode(StatusCodes.Status204NoContent);
--- a/src/VirtoCommerce.CatalogModule.Web/Controllers/Api/CatalogModuleListEntryController.cs
+++ b/src/VirtoCommerce.CatalogModule.Web/Controllers/Api/CatalogModuleListEntryController.cs
@@ -36,16 +36,16 @@ namespace VirtoCommerce.CatalogModule.Web.Controllers.Api
         private readonly ListEntryMover<CatalogProduct> _productMover;
 
         public CatalogModuleListEntryController(
-            IProductIndexedSearchService productIndexedSearchService
-            , ICategoryIndexedSearchService categoryIndexedSearchService
-            , IListEntrySearchService listEntrySearchService
-            , ICategoryService categoryService
-            , IItemService itemService
-            , ICatalogService catalogService
-            , IAuthorizationService authorizationService
-            , ISettingsManager settingsManager
-            , ListEntryMover<Category> categoryMover
-            , ListEntryMover<CatalogProduct> productMover)
+            IProductIndexedSearchService productIndexedSearchService,
+            ICategoryIndexedSearchService categoryIndexedSearchService,
+            IListEntrySearchService listEntrySearchService,
+            ICategoryService categoryService,
+            IItemService itemService,
+            ICatalogService catalogService,
+            IAuthorizationService authorizationService,
+            ISettingsManager settingsManager,
+            ListEntryMover<Category> categoryMover,
+            ListEntryMover<CatalogProduct> productMover)
         {
             _productIndexedSearchService = productIndexedSearchService;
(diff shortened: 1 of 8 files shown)


src/VirtoCommerce.CatalogModule.Web/Controllers/Api/CatalogModuleProductsController.cs
```diff
@@ -282,6 +282,36 @@ public async Task<ActionResult> DeleteProduct([FromQuery] string[] ids)
...
+                var idsToDelete = ids.Skip(i).Take(batchSize);
+                products.AddRange(await _itemsService.GetByIdsAsync(idsToDelete.ToArray(), ItemResponseGroup.ItemInfo.ToString()));
+                var authorizationResult = await _authorizationService.AuthorizeAsync(User, products, new CatalogAuthorizationRequirement(ModuleConstants.Security.Permissions.Delete));
+                if (!authorizationResult.Succeeded)
+                {
+                    return Unauthorized();
+                }
+            }
+
+            for (var i = 0; i < ids.Length; i += batchSize)
```
yecli, line 307: Why do we need the second cycle? Couldn't it be made in one?

src/VirtoCommerce.CatalogModule.Web/Scripts/resources/items.js
```diff
@@ -2,6 +2,7 @@ angular.module('virtoCommerce.catalogModule')
 .factory('virtoCommerce.catalogModule.items', ['$resource', function ($resource) {
     return $resource('api/catalog/products/:id', null, {
         remove: { method: 'DELETE', url: 'api/catalog/products' },
+        plentyRemove: { method: 'POST', url: 'api/catalog/products/plentyDelete' },
```
yecli, line 5: Let's name it `removePlenty`

src/VirtoCommerce.CatalogModule.Web/Controllers/Api/CatalogModuleProductsController.cs
AliveMen: If user haven't permission for one item from the array we throw exception. So we can delete all products only when permission for every product successfully checked.

yecli: Indeed.
I think you should not use `products` here, use `ids` instead in the second cycle.
Products array is not supposed to be changed.

src/VirtoCommerce.CatalogModule.Web/Scripts/blades/categories-items-list.js
```diff
@@ -209,28 +209,31 @@ angular.module('virtoCommerce.catalogModule')
...
-                                    if (blade.mode === 'mappingSource')
-                                        blade.parentBlade.refresh();
-                                }, function (error) {
-                                    bladeNavigationService.setError('Error ' + error.status, blade);
-                                });
+                                listEntries.deletelinks(listEntryLinks,
+                                    () => {
+                                        blade.refresh();
+
+                                        if (blade.mode === 'mappingSource')
```
pushnitsa, line 218: Add braces, please.

pushnitsa (changes requested):

src/VirtoCommerce.CatalogModule.Web/Scripts/resources/items.js
AliveMen: renamed

src/VirtoCommerce.CatalogModule.Web/Controllers/Api/CatalogModuleProductsController.cs
AliveMen: changed

src/VirtoCommerce.CatalogModule.Web/Scripts/blades/categories-items-list.js
AliveMen: done

yecli (approved):

yecli (approved):

src/VirtoCommerce.CatalogModule.Web/Controllers/Api/CatalogModuleListEntryController.cs
```diff
@@ -262,6 +230,50 @@ public async Task<ActionResult> Move([FromBody]ListEntriesMoveRequest moveReques
...
+                return Unauthorized();
+            }
+
+            var idsToDelete = criteria.ObjectIds?.ToList() ?? new List<string>();
+
+            if (idsToDelete.IsNullOrEmpty())
+            {
+                var listEntries = await InnerListItemsSearchAsync(criteria);
+                idsToDelete = listEntries.ListEntries
+                    .Where(x => x.Type.EqualsInvariant(KnownDocumentTypes.Product) || x.Type.EqualsInvariant(KnownDocumentTypes.Category))
```
tatarincev, line 255: use `ProductListEntry.TypeName` and `CategoryListEntry.TypeName` instead of `KnownDocumentTypes`

src/VirtoCommerce.CatalogModule.Web/Controllers/Api/CatalogModuleListEntryController.cs
```diff
@@ -262,6 +230,50 @@ public async Task<ActionResult> Move([FromBody]ListEntriesMoveRequest moveReques
...
+                    .Select(x => x.Id)
+                    .ToList();
+            }
+
+            if (!idsToDelete.IsNullOrEmpty())
+            {
+                for (var i = 0; i < idsToDelete.Count; i++)
```
tatarincev, line 262: I think this for loop will have always N interactions instead of N/page_size

AliveMen: fixed

src/VirtoCommerce.CatalogModule.Web/Controllers/Api/CatalogModuleListEntryController.cs
AliveMen: used

yecli (dismissed):

src/VirtoCommerce.CatalogModule.Web/Controllers/Api/CatalogModuleListEntryController.cs
```diff
@@ -258,10 +228,54 @@ public async Task<ActionResult> Move([FromBody]ListEntriesMoveRequest moveReques
...
+            if (idsToDelete.IsNullOrEmpty())
+            {
+                var listEntries = await InnerListItemsSearchAsync(criteria);
+                idsToDelete = listEntries.ListEntries
+                    .Where(x => x.Type.EqualsInvariant(ProductListEntry.TypeName) || x.Type.EqualsInvariant(CategoryListEntry.TypeName))
+                    .Select(x => x.Id)
+                    .ToList();
+            }
+
+            if (!idsToDelete.IsNullOrEmpty())
```
Vectorfield4, line 262: I think we don't need to check this. Its never be null. If it's empty "for" ends immediately.

Vectorfield4 (approved):

src/VirtoCommerce.CatalogModule.Web/Controllers/Api/CatalogModuleListEntryController.cs
```diff
@@ -16,6 +16,7 @@
 using VirtoCommerce.CatalogModule.Web.Model;
 using VirtoCommerce.Platform.Core.Common;
 using VirtoCommerce.Platform.Core.Settings;
+using VirtoCommerce.SearchModule.Core.Model;
```
Vectorfield4, line 19: Did you need this using?

src/VirtoCommerce.CatalogModule.Web/Controllers/Api/CatalogModuleListEntryController.cs
AliveMen: Agree. Removed.

src/VirtoCommerce.CatalogModule.Web/Controllers/Api/CatalogModuleListEntryController.cs
AliveMen: Removed

yecli (approved):

PR #435, opened 2020-10-08
VP-5470: fix groupby

Vectorfield4:
https://virtocommerce.atlassian.net/browse/VP-5470

### Proposed of changes
Describe the technical details of realization provided by you.

### DIFF
--- a/src/VirtoCommerce.CatalogModule.Data/Repositories/CatalogRepositoryImpl.cs
+++ b/src/VirtoCommerce.CatalogModule.Data/Repositories/CatalogRepositoryImpl.cs
@@ -8,7 +8,6 @@ using Microsoft.EntityFrameworkCore;
 using VirtoCommerce.CatalogModule.Core.Model;
 using VirtoCommerce.CatalogModule.Core.Model.Search;
 using VirtoCommerce.CatalogModule.Data.Model;
-using VirtoCommerce.CoreModule.Core.Seo;
 using VirtoCommerce.Platform.Core.Common;
 using VirtoCommerce.Platform.Data.Extensions;
 using VirtoCommerce.Platform.Data.Infrastructure;
--- a/src/VirtoCommerce.CatalogModule.Data/Repositories/CatalogRepositoryImpl.cs
+++ b/src/VirtoCommerce.CatalogModule.Data/Repositories/CatalogRepositoryImpl.cs
@@ -319,8 +319,8 @@ namespace VirtoCommerce.CatalogModule.Data.Repositories
                     WHERE row_num > 1
                 ";
 
-            var getAllSeoDuplicatesAsyncCommand = CreateCommand(commandTemplate, new string[0]);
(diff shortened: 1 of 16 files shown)


src/VirtoCommerce.CatalogModule.Data/Repositories/CatalogRepositoryImpl.cs
```diff
@@ -303,6 +303,27 @@ public async Task<AssociationEntity[]> GetAssociationsByIdsAsync(string[] associ
             return result;
         }
 
+        public async Task<string[]> GetAllSeoDuplicatesIdsAsync()
+        {
+            const string commandTemplate = @"
+                    WITH cte AS (
```
akak1977, line 309: Why not write there simpler:
``` SQL
SELECT Id FROM 
(
	SELECT
		Id,
		Keyword,
		StoreId,
		count (*) cnt
	FROM CatalogSeoInfo
	GROUP BY 		
		Id,
		Keyword,
		StoreId
) cte
WHERE cte.cnt > 1
```
It's clearer and ANSI-compatible

akak1977 (changes requested):

src/VirtoCommerce.CatalogModule.Data/Repositories/CatalogRepositoryImpl.cs
akak1977: or even (I'm not sure I understood the goal):
``` sql
SELECT DISTINCT CatalogSeoInfo.Id
FROM
(
	SELECT * FROM 
	(
		SELECT
			Keyword,
			StoreId,
			count (*) cnt
		FROM CatalogSeoInfo
		GROUP BY 		
			Keyword,
			StoreId
	) cte
	WHERE cte.cnt > 1
) cte2more
inner join CatalogSeoInfo
on CatalogSeoInfo.Keyword=cte2more.Keyword and CatalogSeoInfo.StoreId=cte2more.StoreId
```

yecli: We are using CTE queries in many places, so I'm not sure we need to avoid them.
Also provided query is simper and more clear for me. 
It is achieving the goal for sure - it returns the ids of duplicated seo infos (by keyword and storeId). So I do not think we need to change this one.
Probably new query does not return id's of only duplicates, but all ids (the first one + duplicates).

yecli (approved):

src/VirtoCommerce.CatalogModule.Data/Repositories/CatalogRepositoryImpl.cs
akak1977: I don't insist.
The second query should return only for duplicates. Does not matter, anyway.

akak1977 (approved):

tatarincev (approved):

yecli (approved):

PR #438, opened 2020-10-21
VP-4146: Add ability to specify alternative text for images

AliveMen:
### Solution
Add migration that extend Image and Asset entities
Add new field to ImageDetail blade
Add new blade for Asset details
Update localisation file

### DIFF
--- a/src/VirtoCommerce.CatalogModule.Core/Model/Asset.cs
+++ b/src/VirtoCommerce.CatalogModule.Core/Model/Asset.cs
@@ -33,10 +33,6 @@ namespace VirtoCommerce.CatalogModule.Core.Model
                         LanguageCode = LanguageCode,
                     }
                 };
-            set
-            {
-                base.SeoInfos = value;
-            }
         }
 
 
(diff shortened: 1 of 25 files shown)


AliveMen:
assets

yecli (approved):

src/VirtoCommerce.CatalogModule.Core/Model/Asset.cs
```diff
@@ -19,6 +21,21 @@ public string ReadableSize
 
         public byte[] BinaryData { get; set; }
 
+        public override IList<SeoInfo> SeoInfos
```
tatarincev, line 24: You don't need to create a SeoInfo with empty values. Move this logic into the  base class and populate specific fields in the derived. 
You can use RelativeURL as slug

AliveMen: fixed

yecli (approved):

yecli (approved):

tatarincev (approved):

PR #450, opened 2020-11-24
Add excluded properties to category

sjroesink:
Adds the ability to store excluded property names per category. The excluded properties are then ignored in the inheritance of a category and its products.

### DIFF
--- a/src/VirtoCommerce.CatalogModule.Core/Extensions/PropertyExtensions.cs
+++ b/src/VirtoCommerce.CatalogModule.Core/Extensions/PropertyExtensions.cs
@@ -1,0 +1,41 @@
+using System.Collections.Generic;
+using System.Linq;
+using VirtoCommerce.CatalogModule.Core.Model;
+using VirtoCommerce.Platform.Core.Common;
+
+namespace VirtoCommerce.CatalogModule.Core.Extensions
+{
+    public static class PropertyExtensions
+    {
+        public static bool HasPropertyExcluded(this IHasExcludedProperties obj, string propertyName)
+        {
+            return obj.ExcludedProperties?.Any(x => x.Name.EqualsInvariant(propertyName)) == true;
+        }
+
+        public static void InheritExcludedProperties<T>(this T obj, IHasExcludedProperties parent)
+            where T : IHasProperties, IHasExcludedProperties
+        {
+            if (parent == null)
+            {
+                return;
+            }
+
+            if (parent.ExcludedProperties?.Any() == true)
+            {
+                // Add the excluded properties as inherited excluded properties
+                var inheritedExcludedProperties = parent.ExcludedProperties.Select(x => x.Inherit()).ToArray();
+                obj.ExcludedProperties ??= new List<ExcludedProperty>();
+                obj.ExcludedProperties.AddDistinct(inheritedExcludedProperties);
+            }
+
+            if (obj.Properties != null && obj.ExcludedProperties?.Any() == true)
+            {
+                var propertiesToRemove = obj.Properties.Where(x => obj.HasPropertyExcluded(x.Name)).ToArray();
+                foreach (var propertyToRemove in propertiesToRemove)
+                {
+                    obj.Properties.Remove(propertyToRemove);
+                }
+            }
+        }
+    }
+}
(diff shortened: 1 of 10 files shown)


src/VirtoCommerce.CatalogModule.Data/Repositories/CatalogDbContext.cs
```diff
@@ -30,6 +32,9 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
                 .WithMany().HasForeignKey(x => x.ParentCategoryId).OnDelete(DeleteBehavior.Restrict);
             modelBuilder.Entity<CategoryEntity>().HasOne(x => x.Catalog)
                 .WithMany().HasForeignKey(x => x.CatalogId).IsRequired().OnDelete(DeleteBehavior.Cascade);
+            modelBuilder.Entity<CategoryEntity>().Property(x => x.ExcludedProperties).HasConversion(
+                x => JsonConvert.SerializeObject(x),
+                x => JsonConvert.DeserializeObject<List<string>>(x));
```
tatarincev, line 37: Are this code is null safe?

src/VirtoCommerce.CatalogModule.Core/Model/Property.cs
```diff
@@ -84,6 +84,15 @@ public virtual bool IsSuitableForValue(PropertyValue propValue)
             return string.Equals(Name, propValue.PropertyName, StringComparison.InvariantCultureIgnoreCase) && ValueType == propValue.ValueType;
         }
 
+        public override bool Equals(object obj)
```
tatarincev, line 87: I would like to recommend to avoid override Equals for Property type in order to avoid side effects that this overriding  may entail

src/VirtoCommerce.CatalogModule.Core/Model/ExcludedProperty.cs
```diff
@@ -0,0 +1,41 @@
+using System;
+using VirtoCommerce.Platform.Core.Common;
+
+namespace VirtoCommerce.CatalogModule.Core.Model
+{
+    public class ExcludedProperty
```
tatarincev, line 6: Better inherit this type from `ValueObject`   then you not need to override any Equals and GetHashCode methods.

src/VirtoCommerce.CatalogModule.Core/Model/CatalogProduct.cs
```diff
@@ -1,260 +1,271 @@
...
+        public virtual void TryInheritFrom(IEntity parent)
+        {
+            this.InheritExcludedProperties(parent as IHasExcludedProperties);
+
+            if (parent is IHasProperties hasProperties)
+            {
+                //Properties inheritance
+                foreach (var parentProperty in hasProperties.Properties ?? Array.Empty<Property>())
+                {
+                    if (this.HasPropertyExcluded(parentProperty.Name))
```
tatarincev, line 178: I  think this line is redundant, all excluded properties have already been  removed after the `InheritExcludedProperties`  call

src/VirtoCommerce.CatalogModule.Core/Model/Property.cs
tatarincev: I think this code also redundant and can be opted-out

src/VirtoCommerce.CatalogModule.Core/Extensions/PropertyExtensions.cs
```diff
@@ -0,0 +1,41 @@
...
+            {
+                // Add the excluded properties as inherited excluded properties
+                var inheritedExcludedProperties = parent.ExcludedProperties.Select(x => x.Inherit()).ToArray();
+                obj.ExcludedProperties ??= new List<ExcludedProperty>();
+                obj.ExcludedProperties.AddDistinct(inheritedExcludedProperties);
+            }
+
+            if (obj.Properties != null && obj.ExcludedProperties?.Any() == true)
+            {
+                var propertiesToRemove = obj.Properties.Where(x => obj.HasPropertyExcluded(x.Name)).ToArray();
```
tatarincev, line 33: You can replace the loop below with this line 
```C#
obj.Properties = obj.Properties.Where(x => !obj.HasPropertyExcluded(x.Name)).ToList()
```

tatarincev: Also, I would like to move these lines out of here because it is breaking the Single responsibility principle

yecli:
Moved this branch into virtocommerce one - https://github.com/VirtoCommerce/vc-module-catalog/tree/feature/VP-6199-category-property-exclusion  - to allow us to make a changes.

PR #462, opened 2020-12-16
VP-6350: Add protection against category hierarchy cycles

yecli:
Moving categories to its children is prohibited now.

### DIFF
--- a/src/VirtoCommerce.CatalogModule.Data/Services/CategoryMover.cs
+++ b/src/VirtoCommerce.CatalogModule.Data/Services/CategoryMover.cs
@@ -34,14 +34,17 @@ namespace VirtoCommerce.CatalogModule.Data.Services
             if (!string.IsNullOrEmpty(moveInfo.Category))
             {
                 var targetCategory = (await _categoryService.GetByIdsAsync(new[] { moveInfo.Category }, CategoryResponseGroup.WithOutlines.ToString())).FirstOrDefault()
-                    ?? throw new InvalidOperationException($"Destination category does not exist");
+                    ?? throw new InvalidOperationException($"Destination category does not exist.");
 
-                foreach (var listEntryOfTypeCategory in moveInfo.ListEntries.Where(listEntry => listEntry.Type.EqualsInvariant(CategoryListEntry.TypeName)))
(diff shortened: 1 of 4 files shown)


src/VirtoCommerce.CatalogModule.Data/Validation/ListEntriesMoveRequestValidator.cs
```diff
@@ -0,0 +1,52 @@
...
+                if (targetCategory == null)
+                {
+                    context.AddFailure($"Destination category does not exist.");
+                    return;
+                }
+
+                foreach (var movedCategory in moveInfo.ListEntries.Where(x => x.Type.EqualsInvariant(CategoryListEntry.TypeName)))
+                {
+                    var movedCategoryPath = string.Join("/", movedCategory.Outline);
+                    var targetCategoryPath = string.Join("/", targetCategory.Outlines.FirstOrDefault());
```
tatarincev, line 36: Why you do not use the `targetCategory.Outline ` here?

src/VirtoCommerce.CatalogModule.Data/Services/CategoryMover.cs
```diff
@@ -51,5 +56,21 @@ public override async Task<List<Category>> PrepareMoveAsync(ListEntriesMoveReque
...
+
+        protected virtual async Task ValidateOperationArguments(ListEntriesMoveRequest moveInfo)
+        {
+            ValidateParameters(moveInfo);
+
+            var validator = new ListEntriesMoveRequestValidator(_categoryService);
+            await validator.ValidateAndThrowAsync(moveInfo);
+        }
+
+        private void ValidateParameters(ListEntriesMoveRequest moveInfo)
```
tatarincev, line 68: I think is better to disable this stupid validation rule

src/VirtoCommerce.CatalogModule.Data/Validation/ListEntriesMoveRequestValidator.cs
yecli: Done, though it increases the complexity of mocking, as this property is calculated with recursion.

src/VirtoCommerce.CatalogModule.Data/Services/CategoryMover.cs
yecli: Done

tatarincev (approved):

PR #466, opened 2020-12-22
#VP-6436 add outline name

kostyrin:
https://virtocommerce.atlassian.net/browse/VP-6436

### DIFF
--- a/src/VirtoCommerce.CatalogModule.Core/Model/OutlinePart/BaseOutlinePartResolver.cs
+++ b/src/VirtoCommerce.CatalogModule.Core/Model/OutlinePart/BaseOutlinePartResolver.cs
@@ -2,10 +2,8 @@ using VirtoCommerce.Platform.Core.Common;
 
 namespace VirtoCommerce.CatalogModule.Core.Model.OutlinePart
 {
-    public abstract class BaseOutlinePartResolver : IOutlinePartResolver
+    public abstract class BaseOutlinePartResolver
     {
-        public abstract string ResolveOutlinePart(IEntity entity);
-
         public string ResolveOutlineName(IEntity entity)
         {
             var result = entity.Id;
--- a/src/VirtoCommerce.CatalogModule.Core/Model/OutlinePart/BaseOutlinePartResolver.cs
+++ b/src/VirtoCommerce.CatalogModule.Core/Model/OutlinePart/BaseOutlinePartResolver.cs
@@ -1,0 +1,24 @@
+using VirtoCommerce.Platform.Core.Common;
+
+namespace VirtoCommerce.CatalogModule.Core.Model.OutlinePart
+{
+    public abstract class BaseOutlinePartResolver : IOutlinePartResolver
+    {
+        public abstract string ResolveOutlinePart(IEntity entity);
+
+        public string ResolveOutlineName(IEntity entity)
+        {
+            var result = entity.Id;
+            if (entity is Catalog catalog)
+            {
+                result = catalog.Name;
+            }
+            else if (entity is Category category)
+            {
+                result = category.Name;
+            }
+
+            return result;
+        }
+    }
+}
--- a/src/VirtoCommerce.CatalogModule.Data/Search/Indexing/AggregationConverter.cs
+++ b/src/VirtoCommerce.CatalogModule.Data/Search/Indexing/AggregationConverter.cs
@@ -239,10 +239,10 @@ namespace VirtoCommerce.CatalogModule.Data.Search.Indexing
                 {
                     return new AggregationItem
                     {
-                        Value = v.Id.Contains(ModuleConstants.Search.OutlineDelimiter) ? v.Id.Split(ModuleConstants.Search.OutlineDelimiter)[0] : v.Id,
+                        Value = v.Id.Contains(ModuleConstants.OutlineDelimiter) ? v.Id.Split(ModuleConstants.OutlineDelimiter)[0] : v.Id,
                         Count = (int)v.Count,
-                        Labels = v.Id.Contains(ModuleConstants.Search.OutlineDelimiter) ?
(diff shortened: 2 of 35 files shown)


src/VirtoCommerce.CatalogModule.Data/Search/Indexing/CategoryDocumentBuilder.cs
```diff
@@ -86,6 +87,11 @@ protected virtual IndexDocument CreateDocument(Category category)
                 document.Add(new IndexDocumentField("__outline", outline.ToLowerInvariant()) { IsRetrievable = true, IsFilterable = true, IsCollection = true });
             }
 
+            foreach (var outlineItem in GetOutlineStrings(category.Outlines, getNameLatestItem: true))
+            {
+                document.Add(new IndexDocumentField($"__outlines", outlineItem) { IsRetrievable = true, IsFilterable = true, IsCollection = true });
```
yecli, line 92: Maybe we could call it:
"__outline_named"
To distinguish from existing `"__outline"` property

src/VirtoCommerce.CatalogModule.Data/Search/Indexing/AggregationConverter.cs
```diff
@@ -235,11 +236,16 @@ protected virtual IList<AggregationItem> GetAttributeAggregationItems(IList<Aggr
         {
             var result = aggregationResponseValues
                 .Select(v =>
-                    new AggregationItem
+                {
+                    return new AggregationItem
                     {
-                        Value = v.Id,
+                        Value = v.Id.Contains('_') ? v.Id.Split('_')[0] : v.Id,
```
yecli, line 242: Having split by '_' here looks a bit dangerous for me, as now any other attribute containing '_' in index value will be changed that way - it could lead to non-working aggregations. Maybe need to limit such parsing appliance somehow.

src/VirtoCommerce.CatalogModule.Data/Search/Indexing/CategoryDocumentBuilder.cs
kostyrin: have added outline delimiter

src/VirtoCommerce.CatalogModule.Data/Search/Indexing/AggregationConverter.cs
kostyrin: have added outline delimiter with '____'

src/VirtoCommerce.CatalogModule.Core/Model/OutlinePart/BaseOutlinePartResolver.cs
```diff
@@ -0,0 +1,22 @@
+using VirtoCommerce.Platform.Core.Common;
+
+namespace VirtoCommerce.CatalogModule.Core.Model.OutlinePart
```
yecli, line 3: In the case of creating a base class probably, you could inherit from `IOutlinePartResolver`, implementing  `ResolveOutlineName` method and leaving `ResolveOutlinePart` abstract.
And probably it should be called `BaseNameOutlinePartResolver`

src/VirtoCommerce.CatalogModule.Data/Search/Indexing/AggregationConverter.cs
```diff
@@ -235,11 +236,16 @@ protected virtual IList<AggregationItem> GetAttributeAggregationItems(IList<Aggr
...
-                    new AggregationItem
+                {
+                    return new AggregationItem
                     {
-                        Value = v.Id,
+                        Value = v.Id.Contains(ModuleConstants.OutlineDelimiter) ? v.Id.Split(ModuleConstants.OutlineDelimiter)[0] : v.Id,
                         Count = (int)v.Count,
-                    })
+                        Labels = v.Id.Contains(ModuleConstants.OutlineDelimiter) ?
+                                    new[] { new AggregationLabel { Label = v.Id.Split(ModuleConstants.OutlineDelimiter)[1] } } :
```
yecli, line 245: Probably you need to check the length to avoid possible IndexOutOfReange exception

src/VirtoCommerce.CatalogModule.Data/Search/Indexing/CategoryDocumentBuilder.cs
yecli: The name `"__outlines"` is still not clear for me, because I do not see a proper semantic difference between `"__outlines"` and `"__outline"` based just on the name. It could be that for `"__outlines"` there are multiple outlines, and for `"__outline"` there is only one. But it is not our case. So maybe `"__outline_named"` would be better

tests/VirtoCommerce.CatalogModule.Tests/CatalogDocumentBuilderTests.cs
```diff
@@ -0,0 +1,133 @@
...
+                        {
+                            Items = new[] {
+                                new OutlineItem { Id = "1", Name = "Catalog" },
+                                new OutlineItem { Id = "11", Name = "Category1" },
+                                new OutlineItem { Id = "12", Name = "Category2" },
+                                new OutlineItem { Id = "111", Name = "Product1" },
+                            } } }, true,
+                        new []
+                        {
+                            $"1/11/12{ModuleConstants.Search.OutlineDelimiter}Category2",
```
yecli, line 88: It could not be compiled, as the constant not in the Search calss

src/VirtoCommerce.CatalogModule.Data/Search/Indexing/CategoryDocumentBuilder.cs
kostyrin: agree

yecli (dismissed):

yecli (approved):

yecli (approved):

PR #468, opened 2020-12-25
VP-6494: Type mismatch in the catalog module

AliveMen:
### Related task:     
   VP-6494 Type mismatch in the catalog module
### Changes:
   Fix return type


yecli (changes requested):
It is a breaking change. We need to change returned value

yecli (approved):

PR #469, opened 2021-01-11
VP-6051: Fix product indexation bug: search names for outline items

akak1977:
That fixes outline items names search. Item ids used lowercased, but by some mistake, ids for names not.


yecli (commented):
Could you please add a test or test case for this fix?

yecli (approved):

PR #480, opened 2021-02-15
VP-6852 Add infinite scroll for catalog and priselist selectors

AliveMen:
### Related task:     
VP-6852: Set new pricelist assignments are only possible for the first 20 catalogues
### Problem:
 We could not use a generic directive for UI scroll because property ObjectIds doesn't work in some *SearchCriteria
### Solution:
Fix search by ObjectIds
### Changes:
Use already existing ObjectIds for added it SearchCriteria the CatalogIds property.

### DIFF
--- a/src/VirtoCommerce.CatalogModule.Core/Model/Search/CatalogSearchCriteria.cs
+++ b/src/VirtoCommerce.CatalogModule.Core/Model/Search/CatalogSearchCriteria.cs
@@ -1,9 +1,14 @@
+using System.Linq;
 using VirtoCommerce.Platform.Core.Common;
 
 namespace VirtoCommerce.CatalogModule.Core.Model.Search
 {
     public class CatalogSearchCriteria : SearchCriteriaBase
     {
-        public string[] CatalogIds { get; set; }
+        public string[] CatalogIds
+        {
+            get => ObjectIds.ToArray();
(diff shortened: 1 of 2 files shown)


src/VirtoCommerce.CatalogModule.Core/Model/Search/CatalogSearchCriteria.cs
```diff
@@ -1,9 +1,14 @@
...
 using VirtoCommerce.Platform.Core.Common;
 
 namespace VirtoCommerce.CatalogModule.Core.Model.Search
 {
     public class CatalogSearchCriteria : SearchCriteriaBase
     {
-        public string[] CatalogIds { get; set; }
+        public string[] CatalogIds
+        {
+            get => ObjectIds.ToArray();
```
eMazeika, line 10: Is NullreferenceException possible here?

akak1977 (dismissed):

kostyrin (dismissed):

src/VirtoCommerce.CatalogModule.Core/Model/Search/CatalogSearchCriteria.cs
AliveMen: fixed

Vectorfield4 (approved):

PR #486, opened 2021-03-09
PT-27: Rebind changed underlying data with ui-grid

eMazeika:
https://virtocommerce.atlassian.net/browse/PT-27


src/VirtoCommerce.CatalogModule.Web/Scripts/blades/item-associations-list.js
```diff
@@ -1,160 +1,162 @@
...
+
+        blade.isLoading = false;
+        blade.refresh = function (item) {
+            initialize(item);
+        };
+
+        function initialize(item) {
+            blade.title = item.name;
+            blade.item = null;
+            $timeout(() => {
```
ksavosteev, line 16: add a comment for why we need to use $timeout here

eMazeika: added

AliveMen (approved):

PR #512, opened 2021-06-07
PT-895: Get languages from the parent catalog

krankenbro:
https://virtocommerce.atlassian.net/browse/PT-895

### DIFF
--- a/src/VirtoCommerce.CatalogModule.Web/Scripts/blades/item-asset-detail.js
+++ b/src/VirtoCommerce.CatalogModule.Web/Scripts/blades/item-asset-detail.js
@@ -14,9 +14,7 @@ angular.module('virtoCommerce.catalogModule')
             $scope.languages = [];
 
             function initialize() {
-                promise.then(function(promiseData) {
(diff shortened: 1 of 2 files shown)


src/VirtoCommerce.CatalogModule.Web/Scripts/blades/item-asset-detail.js
```diff
@@ -14,9 +14,7 @@ angular.module('virtoCommerce.catalogModule')
             $scope.languages = [];
 
             function initialize() {
-                promise.then(function(promiseData) {
-                    $scope.languages = promiseData;
-                });
+                $scope.languages = _.pluck($scope.blade.parentBlade.parentBlade.catalog.languages, 'languageCode');
```
ksavosteev, line 17: I think we should not assume that parentBlade.parentBlade always exists. How about fetching the catalog by Id and taking the list of languages from it?

Vectorfield4 (approved):

ksavosteev (approved):
fix smells please

PR #548, opened 2021-09-14
PT-4315 Products search fix

ksavosteev:
## Description
Products search fix for ProductSearchCriteria.StoreId == null in BrowseFilterService


ksavosteev:
fixed in https://github.com/VirtoCommerce/vc-platform/pull/2372

PR #561, opened 2021-10-14
PT-4224: Extract InnerListItemsSearchAsync to a separate service

ksavosteev:


Vectorfield4 (changes requested):
Create tasks for todos, please.

PR #568, opened 2021-11-03
PT-5268: Delete deleted products from index when indexing by schedule

ksavosteev:

### DIFF
--- a/src/VirtoCommerce.CatalogModule.Data/Services/CatalogService.cs
+++ b/src/VirtoCommerce.CatalogModule.Data/Services/CatalogService.cs
@@ -100,20 +100,20 @@ namespace VirtoCommerce.CatalogModule.Data.Services
                 var catalogs = await GetByIdsAsync(catalogIds);
                 if (!catalogs.IsNullOrEmpty())
                 {
-                    var changedEntriesAggregate = GetDeletedEntries(catalogs);
+                    var changedEntries = GetDeletedEntries(catalogs);
 
-                    await _eventPublisher.Publish(new CatalogChangingEvent(changedEntriesAggregate.CatalogEntries));
-                    await _eventPublisher.Publish(new CategoryChangingEvent(changedEntriesAggregate.CategoryEntries));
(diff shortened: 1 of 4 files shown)


src/VirtoCommerce.CatalogModule.Data/Services/CatalogService.cs
```diff
@@ -184,6 +189,43 @@ private void ClearCache(Catalog[] catalogs)
...
+
+        private CatalogChangedEntriesAggregate GetDeletedEntries(Catalog[] catalogs)
+        {
+            using (var repository = _repositoryFactory())
+            {
+                var catalogChangedEntries = catalogs.Select(x => new GenericChangedEntry<Catalog>(x, EntryState.Deleted)).ToList();
+
+                var foundCatalogIds = catalogs.Select(x => x.Id).ToList();
+                var categoryIds = repository.Categories.Where(x => foundCatalogIds.Contains(x.CatalogId)).Select(x => x.Id).ToList();
+                var productIds = repository.Items.Where(x => foundCatalogIds.Contains(x.CatalogId)).Select(x => x.Id).ToList();
```
Vectorfield4, line 202: Can we make it async?

src/VirtoCommerce.CatalogModule.Data/Services/CatalogService.cs
```diff
@@ -184,6 +189,43 @@ private void ClearCache(Catalog[] catalogs)
...
+                    var entry = new GenericChangedEntry<CatalogProduct>(product, EntryState.Deleted);
+
+                    return entry;
+                }).ToList();
+
+                return new CatalogChangedEntriesAggregate
+                {
+                    CatalogEntries = catalogChangedEntries,
+                    CategoryEntries = categoryChangedEntries,
+                    ProductEntries = productChangedEntries,
```
Vectorfield4, line 226: If we use `CatalogChangedEntriesAggregate` only in the private method and we don`t need to make this aggregate extendable:

1. Change collection type to List<T> instead of IEnumerable<T> (increase performance by removing casts operations);
2. Make `CatalogChangedEntriesAggregate` sealed.

Vectorfield4 (approved):

PR #616, opened 2022-08-04
PT-4249: Catalog/categories filter. There is no applied filter state.

VertyR:
#PT-4249 Catalog/categories filter. There is no applied filter state.

### Jira-link:

https://virtocommerce.atlassian.net/browse/PT-4249


akak1977 (changes requested):
Please, fix smells

akak1977 (approved):

PR #625, opened 2022-08-30
PT-8564: Parent category id is not equal to category id constraint

getmansky:
## Description
Parent category id is not equal to category id constraint


akak1977 (changes requested):
Please use table.UniqueConstraint instead of check constraints

akak1977 (approved):

PR #629, opened 2022-09-16
PT-9059: add null check on convert price range facets

ksavosteev:
## Description
also consolidate platform dependency
### Jira-link:

https://virtocommerce.atlassian.net/browse/PT-9059


akak1977 (approved):

OlegoO (approved):

src/VirtoCommerce.CatalogModule.Data/Search/Indexing/AggregationConverter.cs
```diff
@@ -107,6 +107,11 @@ protected virtual AggregationRequest GetRangeFilterValueAggregationRequest(strin
         protected virtual IList<AggregationRequest> GetPriceRangeFilterAggregationRequests(PriceRangeFilter priceRangeFilter, ProductIndexedSearchCriteria criteria, IList<IFilter> existingFilters)
         {
             var priceRangeFilters = priceRangeFilter.Values?.Select(v => GetPriceRangeFilterValueAggregationRequest(priceRangeFilter, v, existingFilters, criteria.Pricelists)).ToList();
+            if (priceRangeFilters == null)
+            {
+                return null;
```
maikvansark: Its better to return an empty list instead of null. (this also used to be the old return value)

return new List<AggregationRequest>();

PR #675, opened 2023-05-09
PT-12035: Returns Custom Aggregations

OlegoO:
## Description
feat: Returns Custom Aggregations

### Jira-link:
https://virtocommerce.atlassian.net/browse/PT-12035


src/VirtoCommerce.CatalogModule.Data/Search/Indexing/AggregationConverter.cs
```diff
@@ -194,6 +214,17 @@ public virtual async Task<Aggregation[]> ConvertAggregationsAsync(IList<Aggregat
...
 
+        private static void RemoveRange(List<string> terms, string fieldName)
+        {
+            for (var i = terms.Count - 1; i >= 0; i--)
+            {
+                if (terms[i].StartsWith($"{fieldName}-", StringComparison.OrdinalIgnoreCase))
+                {
+                    terms.RemoveAt(i);
+                }
+            }
```
artem-dudarev, line 225: `i` should not be decremented when we remove a term. Otherwise, we are skipping next term.

artem-dudarev (changes requested):

PR #683, opened 2023-05-31
Explicitly state response model for the CatalogModuleProducts_GetProductByIds operation

j-mok:
The workaround for the performance issue in [CatalogModuleProducts_GetProductByIds ](https://github.com/VirtoCommerce/vc-module-catalog/blob/a5ea494738c69ef696dd0ae6fc63560c4ee97c1b/src/VirtoCommerce.CatalogModule.Web/Controllers/Api/CatalogModuleProductsController.cs#L101-L105) operation has the side effect of missing response type in code model resulting in crippled Swagger spec that states just plain 200, without actual response model:
```json
{
    "/api/catalog/products": {
        "get": {
            "tags": [
                "Catalog"
            ],
            "summary": "Gets products by ids",
            "operationId": "CatalogModuleProducts_GetProductByIds",
            "parameters": "...",         
            "responses": {
                "200": {
                    "description": "Success"
                },
                "401": {
                    "description": "Unauthorized"
                },
                "403": {
                    "description": "Forbidden"
                }
            },
            "security": "..."
        }
    }
}
```
This in turn causes client generators to generate plain `Task` or `void` as return type for the operation, effectively preventing clients from consuming the products response.

After the fix, Swagger spec is enriched with response model reference:
```json
{
    "/api/catalog/products": {
        "get": {
            "tags": [
                "Catalog"
            ],
            "summary": "Gets products by ids",
            "operationId": "CatalogModuleProducts_GetProductByIds",
            "parameters": "...",
            "responses": {
                "200": {
                    "description": "Success",
                    "content": {
                        "text/plain": {
                            "schema": {
                                "type": "array",
                                "items": {
                                    "$ref": "#/components/schemas/CatalogProduct"
                                }
                            }
                        },
                        "application/json": {
                            "schema": {
                                "type": "array",
                                "items": {
                                    "$ref": "#/components/schemas/CatalogProduct"
                                }
                            }
                        },
                        "text/json": {
                            "schema": {
                                "type": "array",
                                "items": {
                                    "$ref": "#/components/schemas/CatalogProduct"
                                }
                            }
                        }
                    }
                },
                "401": {
                    "description": "Unauthorized"
                },
                "403": {
                    "description": "Forbidden"
                },
                "404": {
                    "description": "Not Found"
                }
            },
            "security": "..."
        }
    }
}
```

### DIFF
--- a/src/VirtoCommerce.CatalogModule.Web/Controllers/Api/CatalogModuleProductsController.cs
+++ b/src/VirtoCommerce.CatalogModule.Web/Controllers/Api/CatalogModuleProductsController.cs
@@ -86,9 +86,7 @@ namespace VirtoCommerce.CatalogModule.Web.Controllers.Api
         ///<param name="respGroup">Response group.</param>
         [HttpGet]
         [Route("")]
-        [ProducesResponseType(typeof(IEnumerable<CatalogProduct>), StatusCodes.Status200OK)]
-        [ProducesResponseType(typeof(void), StatusCodes.Status404NotFound)]
-        public async Task<ActionResult> GetProductByIds([FromQuery] List<string> ids, [FromQuery] string respGroup = null)
+        public async Task<ActionResult<CatalogProduct[]>> GetProductByIds([FromQuery] List<string> ids, [FromQuery] string respGroup = null)
         {
             var items = await _itemsService.GetAsync(ids, respGroup);
             if (items == null)
--- a/src/VirtoCommerce.CatalogModule.Web/Controllers/Api/CatalogModuleProductsController.cs
+++ b/src/VirtoCommerce.CatalogModule.Web/Controllers/Api/CatalogModuleProductsController.cs
@@ -86,6 +86,8 @@ namespace VirtoCommerce.CatalogModule.Web.Controllers.Api
         ///<param name="respGroup">Response group.</param>
         [HttpGet]
         [Route("")]
+        [ProducesResponseType(typeof(IEnumerable<CatalogProduct>), StatusCodes.Status200OK)]
+        [ProducesResponseType(typeof(void), StatusCodes.Status404NotFound)]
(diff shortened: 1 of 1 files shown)


src/VirtoCommerce.CatalogModule.Web/Controllers/Api/CatalogModuleProductsController.cs
```diff
@@ -86,6 +86,8 @@ public async Task<ActionResult<CatalogProduct[]>> GetByCodes([FromRoute] string
         ///<param name="respGroup">Response group.</param>
         [HttpGet]
         [Route("")]
+        [ProducesResponseType(typeof(IEnumerable<CatalogProduct>), StatusCodes.Status200OK)]
+        [ProducesResponseType(typeof(void), StatusCodes.Status404NotFound)]
         public async Task<ActionResult> GetProductByIds([FromQuery] List<string> ids, [FromQuery] string respGroup = null)
```
artem-dudarev, line 91:
```suggestion
        public async Task<ActionResult<CatalogProduct[]>> GetProductByIds([FromQuery] List<string> ids, [FromQuery] string respGroup = null)
```

artem-dudarev (changes requested):
It's better to return `Task<ActionResult<CatalogProduct[]>>`

src/VirtoCommerce.CatalogModule.Web/Controllers/Api/CatalogModuleProductsController.cs
j-mok: I think the whole point was **not** to return a strongly typed response, but serialize it manually in action's body to avoid performance issue with Newtonsoft.Json.  See the comment I linked in the description and the issue referenced in that comment.

artem-dudarev: The point was to serialize manually, the return type in method signature doesn't matter.

PR #701, opened 2023-10-04
variation image inherits id from parent product image

henrykavalev:
## Description
If requesting a variation product with images that are saved within the parent product, then the id is manually set to null. This PR is setting id to the parent product image id. That way it will return the same image id as if one would request the parent product image id.


henrykavalev:
Since the other parameters were inherited from parentAssetBase, it led on our side to the question why it is not done for the id as well. For us it is easier to differentiate between pictures based on the id instead of the name. For example working on frontend with Keys of the HTML Lists Items, where we could use Id as the Key to avoid duplicated image names.

henrykavalev:
Hello together, since there are no updates on this issue for some time now, I wanted to ask whether this issue will be implemented or if I can help in any other way regarding this suggestion?

PR #711, opened 2023-11-21
Fix: Possible to enter Min. quantity greater than Max. quantity, pord…

cgenesoni01:
…uct details.

## Description
Implement validation for product details, ensuring that the minimum quantity is less than or equal to the maximum quantity.

### DIFF
--- a/src/VirtoCommerce.CatalogModule.Web/Scripts/blades/item-detail.js
+++ b/src/VirtoCommerce.CatalogModule.Web/Scripts/blades/item-detail.js
@@ -65,9 +65,18 @@ angular.module('virtoCommerce.catalogModule')
             return !angular.equals(blade.item, blade.origItem) && blade.hasUpdatePermission();
         }
 
-        function canSave() {
-            return isDirty() && blade.formScope && blade.formScope.$valid;
+        function isValidQuantity(item) {
+            let minQuantity = parseInt(item.minQuantity, 10);
+            let maxQuantity = parseInt(item.maxQuantity, 10);
+            
+            return !(isNaN(minQuantity) || isNaN(maxQuantity) || minQuantity > maxQuantity);
         }
+
+        function canSave() {
+            return isDirty() && isValidQuantity(blade.item) && blade.formScope && blade.formScope.$valid;
+        }
+
+        
 
         function saveChanges() {
             blade.isLoading = true;


src/VirtoCommerce.CatalogModule.Web/Scripts/blades/item-detail.js
```diff
@@ -65,10 +65,19 @@ angular.module('virtoCommerce.catalogModule')
             return !angular.equals(blade.item, blade.origItem) && blade.hasUpdatePermission();
         }
 
+        function isValidQuantity(item) {
```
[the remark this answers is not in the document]
cgenesoni01: Resolved the issue when the maxQuantity had no values.

src/VirtoCommerce.CatalogModule.Web/Scripts/blades/item-detail.js
```diff
@@ -65,10 +65,31 @@ angular.module('virtoCommerce.catalogModule')
             return !angular.equals(blade.item, blade.origItem) && blade.hasUpdatePermission();
         }
 
+        function isEmptyValue(value) {
+            return value == null || value === '';
+        }
+
+        function isValidQuantity(item) {
+
```
tatarincev, line 73: item.minQuantity and item.maxQuantity both can be null or not defined, please also consider this requirement.

cgenesoni01: IsEmptyValue is consider undefined and null cases, if you test it with null or undefined you can see that the method IsValidQuantity is returning true

src/VirtoCommerce.CatalogModule.Web/Scripts/blades/item-detail.js
```diff
@@ -65,10 +65,31 @@ angular.module('virtoCommerce.catalogModule')
...
+            if (isEmptyValue(item.minQuantity) || isEmptyValue(item.maxQuantity)) {
+                return true;
+            }
+
+            let minQuantity = parseInt(item.minQuantity, 10);
+            let maxQuantity = parseInt(item.maxQuantity, 10);
+
+            if (isNaN(maxQuantity) || isNaN(minQuantity)) {
+                return true;
+            }
```
artem-dudarev, line 83: If entered value is not a number, then it's not valid

artem-dudarev (changes requested):

src/VirtoCommerce.CatalogModule.Web/Scripts/blades/item-detail.js
artem-dudarev: What about whitespaces? `' '`

artem-dudarev (approved):

PR #730, opened 2024-05-08
Ensure that saved editorial reviews are not marked as inherited

j-mok:
When I have a product variation that inherit reviews (descriptions) from the parent, these reviews are editable, but the edits won't get persisted in the database - they get silently dropped without a warning. This behavior is due to the fact that those edited inherited reviews are ignored when mapping `CatalogProduct` to `ItemService` - [there's a filter](https://github.com/VirtoCommerce/vc-module-catalog/blob/f55012d8b82928a681dc1b3d2ef01b44ff818beb/src/VirtoCommerce.CatalogModule.Data/Model/ItemEntity.cs#L356) on `EditorialReview.IsInherited` property.

This PR offers a small change in the UI code, that clears the `isInherited` flag on those editorial reviews that got saved, effectively promoting them to variation's own reviews. In effect, the user can now directly override inherited reviews in variants.

### DIFF
--- a/src/VirtoCommerce.CatalogModule.Web/Scripts/blades/editorialReview-detail.js
+++ b/src/VirtoCommerce.CatalogModule.Web/Scripts/blades/editorialReview-detail.js
@@ -28,8 +28,9 @@ angular.module('virtoCommerce.catalogModule')
             var existReview = _.find(blade.item.reviews, function (x) { return x == blade.origEntity; });
             if (!existReview) {
                 blade.item.reviews.push(blade.origEntity);
-            };
+            }
             angular.copy(blade.currentEntity, blade.origEntity);
+            blade.currentEntity.isInherited = false;
             $scope.bladeClose();
         };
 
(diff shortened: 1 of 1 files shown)


OlegoO (approved):

src/VirtoCommerce.CatalogModule.Web/Scripts/blades/editorialReview-detail.js
```diff
@@ -28,8 +28,9 @@ angular.module('virtoCommerce.catalogModule')
             var existReview = _.find(blade.item.reviews, function (x) { return x == blade.origEntity; });
             if (!existReview) {
                 blade.item.reviews.push(blade.origEntity);
-            };
+            }
             angular.copy(blade.currentEntity, blade.origEntity);
+            blade.currentEntity.isInherited = false;
```
artem-dudarev, line 33: Please remove code duplication by calling the `saveChanges()` function, which is declared later in this file.

src/VirtoCommerce.CatalogModule.Web/Scripts/blades/editorialReview-detail.js
```diff
@@ -102,9 +103,10 @@ angular.module('virtoCommerce.catalogModule')
             var existReview = _.find(blade.item.reviews, function (x) { return x === blade.origEntity; });
             if (!existReview) {
                 blade.item.reviews.push(blade.origEntity);
-            };
+            }
 
             angular.copy(blade.currentEntity, blade.origEntity);
+            blade.currentEntity.isInherited = false;
```
artem-dudarev, line 109: This flag should be changed before `angular.copy()`. Now it doesn't have any effect.

artem-dudarev (changes requested):

artem-dudarev (approved):

PR #759, opened 2024-12-03
feat: make some methods of BrowseFilterService protected

dvs39:
## Description
Minor refactoring, extend BrowseFilterService for reuse

### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/VirtoCommerce.Catalog_3.833.0-pr-759-ff2f.zip


src/VirtoCommerce.CatalogModule.Core/Model/AggregationProperty.cs
```diff
@@ -1,24 +1,24 @@
...
-            result.Values = Values?.ToList();
-            return result;
-        }
-    }
-}
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace VirtoCommerce.CatalogModule.Web.Model
```
alexeyshibanov, line 5: Move to correct namespace

artem-dudarev (approved):

PR #833, opened 2025-09-24
VCST-3911: Remove leading and trailing spaces from product name

ghost:
## Description
- Added the rules for cleaning the product name from tab characters and line breaks have been changed.
### Jira-link:
https://virtocommerce.atlassian.net/browse/VCST-3911
### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/VirtoCommerce.Catalog_3.896.0-pr-833-f8a7.zip


src/VirtoCommerce.CatalogModule.Web/Controllers/Api/CatalogModuleProductsController.cs
```diff
@@ -1,6 +1,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Text.RegularExpressions;
```
artem-dudarev, line 4: Remove redundant `using` and configure you IDE to remove them automatically

PR #882, opened 2026-05-07
Fix: cascade CatalogId to descendants and products on cross-catalog category move (cut & paste)

yuskithedeveloper:
### Problem

Moving a category to a different catalog via the admin UI's Cut/Paste
(`POST /api/catalog/listentries/move`) leaves the database in an
inconsistent state: the selected `Category.CatalogId` is updated, but
nothing else is. Specifically:

- **Subcategories** keep their original `CatalogId`, so an entire
  subtree ends up belonging to a different catalog than its root.
- **Products** under the moved subtree keep their original `CatalogId`,
  so `Item.CatalogId` no longer matches `Category.CatalogId`.

This breaks any feature that filters by `CatalogId` (search, indexing,
catalog-scoped property/dictionary lookups).

### Cause

`CategoryMover.PrepareMoveAsync` only iterates entries explicitly
listed in `ListEntriesMoveRequest.ListEntries` (the user's selection
in the UI) and mutates `CatalogId`/`ParentId` on those rows alone.
There is no traversal of the moved subtree, and `ProductMover` only
touches products that were directly listed in the request — products
located *inside* a moved category are invisible to it.

### Fix

`CategoryMover` now cascades the catalog change. The change is
contained to a single file; controller, request DTO, and
`ProductMover` are unchanged.

- **`PrepareMoveAsync`** captures the source `CatalogId` of each
  cross-catalog root, then loads each source catalog's category set
  once via `ICategorySearchService` and BFS-walks a parent→children
  map to collect all descendants. Descendants get the target
  `CatalogId` and are appended to the returned list.
- **`ConfirmMoveAsync`** saves all categories (roots + descendants)
  and then cascades products. The target `CatalogId` and category ids
  are derived from the entities themselves — no instance fields, no
  changes to the base mover signature.
- **`CascadeProductsAsync`** pages through products by `CategoryIds`
  with `SearchInVariations = true` so variations are updated as flat
  results, then saves each batch via `IItemService.SaveChangesAsync`
  (which raises the existing domain events that trigger reindexing).

Same-catalog reparents are unchanged in behavior: no descendants are
added in `Prepare`, and the product cascade in `Confirm` is a single
bounded search whose per-product `CatalogId` filter yields no writes.

### Behavior considerations

- `CategoryMover` gains three new dependencies (`ICategorySearchService`,
  `IProductSearchService`, `IItemService`). DI resolves them
  automatically; no module wiring change is needed.
- For very large catalogs, a cross-catalog move now does proportionally
  more work inside the HTTP request. The existing endpoint stays
  synchronous; converting moves to a background job is left as a
  follow-up.
- Indexing is not triggered synchronously — `IItemService.SaveChangesAsync`
  raises the usual `CatalogProductChangedEvent` and reindexing happens
  through the existing pipeline.

### Artifact URL:


src/VirtoCommerce.CatalogModule.Data/Services/CategoryMover.cs
```diff
@@ -67,5 +114,138 @@ protected virtual async Task ValidateOperationArguments(ListEntriesMoveRequest m
...
+                {
+                    output.Add(child);
+                    queue.Enqueue(child.Id);
+                }
+            }
+        }
+
+        private async Task CascadeProductsAsync(IList<string> categoryIds, string targetCatalogId)
+        {
+            foreach (var idsChunk in categoryIds.Chunk(ProductPageSize))
```
DmitryGrishinVirtoworks: I think logically variable that we use in these places are different (event if we need the save value for them)

categoryIds.Chunk(ProductPageSize)
criteria.Take = ProductPageSize;

I think it would be better to create new constant for first place, something like CategoryIdsBatchSize

src/VirtoCommerce.CatalogModule.Data/Services/CategoryMover.cs
```diff
@@ -67,5 +114,138 @@ protected virtual async Task ValidateOperationArguments(ListEntriesMoveRequest m
...
+            return result.Results;
+        }
+
+        private static void AddToParentMap(Dictionary<string, List<Category>> byParent, Category category)
+        {
+            var parentKey = category.ParentId ?? string.Empty;
+            if (!byParent.TryGetValue(parentKey, out var children))
+            {
+                byParent[parentKey] = children = new List<Category>();
+            }
```
DmitryGrishinVirtoworks, line 182: It would be easier to read if we separate if block from children.Add(category); with new line

src/VirtoCommerce.CatalogModule.Data/Services/CategoryMover.cs
```diff
@@ -67,5 +114,138 @@ protected virtual async Task ValidateOperationArguments(ListEntriesMoveRequest m
...
+                    if (toSave.Count > 0)
+                    {
+                        await _itemService.SaveChangesAsync(toSave.ToArray());
+                    }
+
+                    if (page.Results.Count < ProductPageSize)
+                    {
+                        break;
+                    }
+                    skip += ProductPageSize;
```
DmitryGrishinVirtoworks, line 250: It would be easier to read if we separate if block from skip += ProductPageSize; with new line

src/VirtoCommerce.CatalogModule.Data/Services/CategoryMover.cs
```diff
@@ -13,22 +15,45 @@ namespace VirtoCommerce.CatalogModule.Data.Services
...
+            }
+
+            await _categoryService.SaveChangesAsync(categories.ToArray());
+
+            // Cascade CatalogId to products that live under any of the moved categories.
+            // Every entity here carries the target CatalogId (set in PrepareMoveAsync), and the
+            // per-product filter in CascadeProductsAsync makes same-catalog reparents a cheap no-op.
+            var targetCatalogId = categories[0].CatalogId;
+            var categoryIds = categories.Select(x => x.Id).ToArray();
+            await CascadeProductsAsync(categoryIds, targetCatalogId);
```
DmitryGrishinVirtoworks, line 58:
```suggestion
            
            await CascadeProductsAsync(categoryIds, targetCatalogId);
```

DmitryGrishinVirtoworks (commented):
logic looks good, just take a look my comments please

src/VirtoCommerce.CatalogModule.Data/Services/CategoryMover.cs
yuskithedeveloper: Used `CategoryIdsChunkSize` ("Chunk" to avoid confiusion with criteria page size)

DmitryGrishinVirtoworks (approved):

PR #894, opened 2026-06-23
VCST-5177: Configurable products sorting

yuskithedeveloper:
## Description
Introduces store-configurable "Sort by" options for category browsing and search, replacing the previously hardcoded sort list. Sort orderings are defined in code as built-ins, and store admins can now rename, reorder, hide/show, and add their own custom orderings — managed from Store → Search configuration → Sorting. Only the admin's changes are stored (as overrides), so built-ins keep working out of the box with no setup. Includes the admin UI and unit tests.
image
### Sort-ordering badges (backend admin UI) 
| Badge | Style | Shown when | What it means |
|-------|-------|-----------|----------------|
| **DEFAULT** | green pill | the ordering is the current default | It's the **first visible** ordering in the list. This is the order customers get when they haven't picked a sort (empty `?sort`). It's computed from position, not stored — drag a different ordering to the top and DEFAULT moves to it. Exactly one row has it. |
| **HIDDEN** | solid grey pill | the ordering's **Visible** toggle is off | The option is **not shown** in the storefront "Sort by" dropdown. It still exists and can be referenced by its code, but customers can't pick it from the UI. (Hidden rows are also dimmed.) |
| **CUSTOM** | grey outline pill | the ordering was created by an admin | An **admin-authored** ordering (not one of the built-in ones that ship in code). Only CUSTOM rows can be deleted — they show the **✕** icon. Built-ins can be hidden but not deleted. |
| **READ-ONLY** | grey outline pill | the ordering's expression is locked | The **sort clauses can't be edited** — the expression is fixed/computed in code. In the detail blade the clause editor is shown but disabled. (You can still rename / reorder / hide it unless otherwise restricted.) |
 
- **Combinations are normal.** E.g. a built-in ordering at the top that's code-locked shows **DEFAULT + READ-ONLY**; an admin ordering that's hidden shows **CUSTOM + HIDDEN**.
- **DEFAULT follows visibility:** if you hide the current default, DEFAULT jumps to the next visible ordering down the list.
- The labels are localized (`catalog.blades.product-sorting-list.badges.*`), so in non-English locales the text changes but the meaning/colors stay the same.

### Extensibility C# code examples

VirtoCommerce.XCart.Web below acts as a custom module (assembly) where the extension lives

```
using System;
using System.Linq;
using Microsoft.Extensions.DependencyInjection;
using VirtoCommerce.CatalogModule.Core.Search.Sorting;
using VirtoCommerce.CatalogModule.Core.Search.Sorting.Resolvers;
namespace VirtoCommerce.XCart.Web;
// ===========================================================================================================
// VCST-5177 — code-only sorting extension EXAMPLES (TEMPORARY, hosted in XCart for testing).
// Three things you can do from code, with NO admin UI action and NO DB seeding/migration:
//   1) add a new built-in ordering        -> VendorAscendingProductSortingResolver
//   2) remove built-in ordering(s)         -> AddProductSortingExamples() removes the two date orderings
//   3) a complex, request-conditional one  -> SmartProductSortingResolver
// Wire-up is a single call from Module.Initialize: serviceCollection.AddProductSortingExamples();
// To revert: delete this file and that one line.
// ===========================================================================================================
/// <summary>
/// EXAMPLE 1 — a new built-in ordering "Vendor (A-Z)".
/// NOTE: `vendor` must be an indexed, sortable field on the product document. If your index doesn't have it,
/// swap the expression for a field that does (e.g. `availability:desc`).
/// </summary>
public class VendorAscendingProductSortingResolver : AbstractExpressionProductSortingResolver
{
    public override string Code => "vendor-ascending";              // stable key: API + ?sort=<code>
    public override ProductSortingInfo Info { get; } = new()
    {
        Name = "Vendor (A-Z)",
        Order = 35,                                                 // position in the dropdown (not first → not the default)
        IsVisible = true,
        // AllowOverride / IsExpressionEditable default to true → the admin may rename/reorder/hide & edit its clauses.
    };
    protected override string DefaultExpression => "vendor:asc";    // logical expr; bound to the physical index field downstream
}
/// <summary>
/// EXAMPLE 3 — a complex ordering whose expression adapts to the current request (the category being browsed),
/// recomputed per request from the context. Here: in "Bolts" sort A→Z, in "Snacks" sort Z→A, otherwise relevance.
/// </summary>
public class SmartProductSortingResolver : IProductSortingResolver
{
    // Real B2B-store category ids (Catalog.Category table).
    private const string BoltsCategoryId = "02fe37dcaeb2458a831011abe43fd335";
    private const string SnacksCategoryId = "532a6b5a-cf15-461a-836e-71bad60d49a3";
    public string Code => "smart";
    public ProductSortingInfo Info { get; } = new()
    {
        Name = "Recommended",
        Order = 10,                     // set to 0 to make it the store default (first visible)
        IsVisible = true,
        IsExpressionEditable = false,   // computed: admin can rename/hide, but the clause editor stays read-only
    };
    public string GetSortExpression(ProductSortingContext context)
    {
        // (a) Browsing "Bolts" → alphabetical A→Z.
        if (IsInCategory(context, BoltsCategoryId))
        {
            return "name:asc";
        }
        // (b) Browsing "Snacks" → alphabetical Z→A.
        if (IsInCategory(context, SnacksCategoryId))
        {
            return "name:desc";
        }
        // (c) Anywhere else → a single, simple clause (one clause is enough; clauses are required but the count is free).
        return "id:asc";
    }
    // The X-API populates CategoryId and Outline on the context (from the current browse request), so we match on those
    // first. We also fall back to a substring match on the raw Filter (the category is embedded as
    // `category.subtree:<catalogId>/<categoryId>`): category ids are unique GUIDs, so the match is safe and also covers
    // descendants (a child's subtree outline contains the ancestor id).
    private static bool IsInCategory(ProductSortingContext context, string categoryId) =>
        context.CategoryId == categoryId ||
        context.Outline?.Contains(categoryId) == true ||
        context.Filter?.Contains(categoryId, StringComparison.OrdinalIgnoreCase) == true;
}
public static class ProductSortingExampleExtensions
{
    public static IServiceCollection AddProductSortingExamples(this IServiceCollection services)
    {
        // EXAMPLE 1 — contribute a new ordering (one line per resolver).
        services.AddSingleton<IProductSortingResolver, VendorAscendingProductSortingResolver>();
        // EXAMPLE 3 — contribute the request-conditional ordering.
        services.AddSingleton<IProductSortingResolver, SmartProductSortingResolver>();
        // EXAMPLE 2 — remove built-in ordering(s): drop both date orderings entirely (gone from the dropdown AND the
        // API — not even resolvable by code). This works because XCart's manifest depends on Catalog, so Catalog has
        // already registered the built-ins by the time XCart.Initialize runs.
        var toRemove = services
            .Where(d => d.ServiceType == typeof(IProductSortingResolver) &&
                        (d.ImplementationType == typeof(CreatedDateAscendingProductSortingResolver) ||
                         d.ImplementationType == typeof(CreatedDateDescendingProductSortingResolver)))
            .ToList();
        foreach (var descriptor in toRemove)
        {
            services.Remove(descriptor);
        }
        return services;
    }
}
```

### Jira-link:
https://virtocommerce.atlassian.net/browse/VCST-5177
### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/VirtoCommerce.Catalog_3.1032.0-alpha.2558-vcst-5177-configurable-products-sorting.zip


OlegoO (changes requested):
1. Need to add the setting name and description into the localisation files
2. Update or create a readme with information on how it works

src/VirtoCommerce.CatalogModule.Core/Search/Sorting/Resolvers/FeaturedProductSortingResolver.cs
```diff
@@ -0,0 +1,15 @@
...
+/// <c>priority</c> is a logical field; the search-request builder binds it to the (per-category) priority index
+/// field(s).
+/// </summary>
+public class FeaturedProductSortingResolver : AbstractExpressionProductSortingResolver
+{
+    public override string Code => ProductSortingConsts.Featured;
+
+    public override ProductSortingInfo Info { get; } = new() { Name = "Featured", Order = ProductSortingConsts.FeaturedOrder };
+
+    protected override string DefaultExpression => "__score:desc;priority:desc;id:asc";
```
alexeyshibanov, line 14: The doc-comment above (and `AbstractExpressionProductSortingResolver`) states the search-request builder binds `priority` to the per-category priority index field(s). But the x-catalog `IndexSearchRequestBuilder.AddSorting` has no `priority` case — it falls through `default:` and sorts on the flat `priority` field, so the browsed category is ignored (`priority_{catalogId}_{categoryId}` is never used). Per-category Featured ordering isn't actually implemented, even though `ProductSortingContext.CategoryId`/`Outline` are plumbed through.

Separately, please confirm this is an intended behaviour change: the previous empty-sort default was relevance only, whereas Featured now appends `priority:desc;id:asc`, which reorders existing stores' default browse/search output.

Either add a per-category `priority` binding in the builder, or adjust the comments to match the flat-priority behaviour.

src/VirtoCommerce.CatalogModule.Core/Search/Sorting/ProductSortingInfo.cs
```diff
@@ -0,0 +1,37 @@
+namespace VirtoCommerce.CatalogModule.Core.Search.Sorting;
+
+/// <summary>
+/// The code-provided defaults for a sorting. A resolver supplies concrete values (no nulls);
+/// the store admin may override them at runtime depending on the two gate flags below.
+/// </summary>
+public class ProductSortingInfo
```
alexeyshibanov, line 7: Minor hardening: resolvers expose `Info` as a shared singleton instance and this type has public setters. It's read-only in practice today, but since resolvers are registered as singletons, any future in-place mutation of `Info` would be process-wide shared state. Consider making these properties init-only.

alexeyshibanov (commented):
Two notes: one on the documented vs. actual `priority` binding for the Featured ordering (plus a default-order behaviour change to confirm), and one minor hardening suggestion.

alexeyshibanov (changes requested):

src/VirtoCommerce.CatalogModule.Web/Scripts/blades/product-sorting-detail.tpl.html
```diff
@@ -0,0 +1,85 @@
...
+                <input type="checkbox" ng-model="blade.entity.isVisible" ng-disabled="!blade.canEditDisplay">
+                <span class="switch"></span>
+              </label>
+            </div>
+          </div>
+
+          <div class="form-group">
+            <label class="form-label required">{{ 'catalog.blades.product-sorting-detail.labels.name' | translate }}</label>
+            <div class="form-input">
+              <input class="form-control" type="text" ng-model="blade.entity.name" ng-change="blade.onNameChanged()" ng-disabled="!blade.canEditDisplay" required>
```
alexeyshibanov: Consistency: these inputs (and the clause `<select>` further down) use `class=form-control`, but the sibling editor reached from the same Search configuration blade — `aggregation-properties-details.tpl.html` (Facets) — and the platform canonical (`role-detail`) use a **bare** `<input ng-model>` directly under `.form-input`. The admin theme styles `.form-input > input`; `form-control` opts out of that, so the two editors in the same area can render inconsistently. Suggest dropping `form-control` here (lines 20, 27, 40 and the clause select) to match the Facets blade.

Separately: the form is `name=formScope` with `required` attributes, but `canExecuteMethod` uses a hand-rolled `isValid()` instead of `formScope.$valid` — minor, the Angular form state is already there.

src/VirtoCommerce.CatalogModule.Web/Scripts/blades/product-sorting-list.js
```diff
@@ -0,0 +1,168 @@
...
+            productSortings.saveSortings({ storeId: blade.storeId }, blade.currentEntities, function () {
+                closeChildren();
+                initializeBlade();
+            }, function (error) {
+                blade.isLoading = false;
+                bladeNavigationService.setError('Error ' + error.status, blade);
+            });
+        };
+
+        function closeChildren() {
```
alexeyshibanov: Minor: `closeChildren()` re-implements `bladeNavigationService.closeChildrenBlades(blade, callback)`, the platform helper for exactly this. Reusing it keeps child-blade closing consistent with the rest of the admin and drops the hand-rolled loop.

src/VirtoCommerce.CatalogModule.Web/Scripts/blades/search-configuration.js
```diff
@@ -0,0 +1,34 @@
...
+            template: 'Modules/$(VirtoCommerce.Catalog)/Scripts/blades/product-sorting-list.tpl.html'
+        }, blade);
+    }
+
+    function openFacets() {
+        // Reuse the existing filtering-properties (facets) blade unchanged.
+        bladeNavigationService.showBlade({
+            id: 'storeFilteringProperties',
+            storeId: blade.storeId,
+            title: 'Filtering properties',
```
alexeyshibanov: Nit: this title is a hard-coded English string while every other title/subtitle in this blade uses a translate key. Consider a proper localization key (you already have `catalog.blades.search-configuration.menu.facets`) so the Facets blade title is localized too.

alexeyshibanov (commented):
Admin UI (AngularJS) notes. Overall the blades follow the master-detail draft pattern well (in-memory edits in the list, single persist on Save; empty-state and list are mutually exclusive; delete is gated to custom orderings and closes children; the widget is `isVisible: !isNew`; the old aggregation widget is removed cleanly with no orphan references). A few consistency/nit items below.

src/VirtoCommerce.CatalogModule.Core/Search/Sorting/Resolvers/FeaturedProductSortingResolver.cs
yuskithedeveloper: Implemented per-category priority support in XCatalog.

PR #904, opened 2026-08-07
VCST-5387: Stream catalog binaries during backup and restore

ToxaKZ:
## Description

Catalog image and file binaries are no longer embedded as base64 in catalog JSON when `HandleBinaryData` is enabled.

With a sidecar-aware export orchestrator, the Catalog module now keeps `VirtoCommerce.Catalog.json` as normal, readable JSON and streams binaries to top-level backup entries under their source paths, for example:

- `VirtoCommerce.Catalog.json`
- `assets/catalog/products/image.jpeg`
- `assets/catalog/products/manual.pdf`

The JSON stores relative `binaryDataReference` values that map each category, product, or variation image/asset to its top-level ZIP entry. Source folder hierarchy, file names, and real extensions are preserved so a backup can be inspected directly. Duplicate source URLs reuse one entry, while external URLs remain unchanged and are not copied into the backup.

During restore, Catalog reads top-level sidecars through the Platform import contract and streams them directly to blob storage after their entities are saved. References are validated against destination URLs before a blob is written. Missing source files during export are logged as warnings; the affected entity is still exported without binary data and the module export continues successfully.

Compatibility is preserved in both directions:

- older orchestrators still use the legacy single-stream contract and receive the versioned nested Catalog package;
- previously produced nested packages using either readable source paths or SHA-256 `.bin` references remain importable;
- legacy plain JSON backups with inline base64 binary data remain importable;
- with `HandleBinaryData` disabled, export remains plain JSON and binary content is not restored.

The implementation streams asset content and uses temporary files only where a seekable module payload is required; it does not accumulate the backup or all asset data in memory. Platform dependencies and the module manifest target the released Platform `3.1062.0` that provides the optional sidecar contracts. The Catalog module manifest requires BackupRestore `3.1005.0` so the compatible top-level binary orchestrator is installed.

### QA-test:
1. Install Platform `3.1062.0`, BackupRestore `3.1005.0`, and the Catalog build from this PR.
2. In Manager UI, upload an image or file asset and attach it to a category, product, or variation.
3. Open **Backup and restore** and start a backup with **Binary** and **VirtoCommerce.Catalog** selected.
4. Open the resulting ZIP and verify that `VirtoCommerce.Catalog.json` is readable JSON and the uploaded file exists under `assets/<sourceUrl>` with its original name and extension. Verify that no `assets/*.bin` entry is created for the new format.
5. Verify that the JSON contains a matching `binaryDataReference` and does not contain inline base64 `binaryData`.
6. Delete the source asset through Manager UI, restore the backup, and verify that the file and its catalog association are restored with matching size/checksum.
7. Keep a catalog reference whose physical file is missing and repeat export. The backup must complete successfully and the missing binary must be logged as a warning rather than an export error.
8. Export a catalog without images/assets and verify that the backup remains valid and contains no Catalog sidecar entries.
9. Restore a backup produced before this change with binary handling enabled and verify that legacy nested assets are restored.
10. Export and import with **Binary** disabled and verify that the Catalog module part remains plain JSON and no binary content is restored.

### Jira-link:
https://virtocommerce.atlassian.net/browse/VCST-5387

### Artifact URL:
https://vc3prerelease.blob.core.windows.net/packages/VirtoCommerce.Catalog_3.1042.0-pr-904-2581.zip

### DIFF
--- a/src/VirtoCommerce.CatalogModule.Core/Model/AssetBase.cs
+++ b/src/VirtoCommerce.CatalogModule.Core/Model/AssetBase.cs
@@ -19,6 +19,12 @@ namespace VirtoCommerce.CatalogModule.Core.Model
         }
         public string RelativeUrl { get; set; }
         public string Url { get; set; }
+        /// <summary>
+        /// Gets or sets the transient relative reference to binary data in a catalog backup package.
+        /// The property is populated only while a backup is exported or restored and is not persisted.
+        /// </summary>
+        [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
+        public string BinaryDataReference { get; set; }
         public string Description { get; set; }
         public int SortOrder { get; set; }
         [JsonIgnore]
--- a/src/VirtoCommerce.CatalogModule.Data/ExportImport/CatalogExportImport.cs
+++ b/src/VirtoCommerce.CatalogModule.Data/ExportImport/CatalogExportImport.cs
@@ -53,8 +53,8 @@ namespace VirtoCommerce.CatalogModule.Data.ExportImport
         {
             public ExportImportOptions Options { get; init; }
             public CatalogImportPackage Package { get; init; }
-            public IDictionary<string, IList<ProductAssociation>> AssociationBackupMap { get; init; }
-            public ISet<string> AlreadySavedIds { get; init; }
+            public Dictionary<string, IList<ProductAssociation>> AssociationBackupMap { get; init; }
+            public HashSet<string> AlreadySavedIds { get; init; }
             public ImportStageContext ProductsStage { get; init; }
             public ImportStageContext VariationsStage { get; init; }
             public ExportImportProgressInfo ProgressInfo { get; init; }
@@ -532,7 +532,7 @@ namespace VirtoCommerce.CatalogModule.Data.ExportImport
 
         private static void AddCategoryByHierarchyLevel(
             Category category,
-            ICollection<Category> rootCategories,
+            List<Category> rootCategories,
             IDictionary<int, IList<Category>> categoriesByHierarchyLevel)
         {
             if (category.Level <= 0)
@@ -728,8 +728,8 @@ namespace VirtoCommerce.CatalogModule.Data.ExportImport
 
         private static (List<CatalogProduct> Parents, List<CatalogProduct> Variations) PrepareProducts(
             IEnumerable<CatalogProduct> products,
-            IDictionary<string, IList<ProductAssociation>> associationBackupMap,
-            ISet<string> alreadySavedIds)
+            Dictionary<string, IList<ProductAssociation>> associationBackupMap,
+            HashSet<string> alreadySavedIds)
         {
             var parentsToSave = new List<CatalogProduct>();
             var variationsToSave = new List<CatalogProduct>();
@@ -763,7 +763,7 @@ namespace VirtoCommerce.CatalogModule.Data.ExportImport
             return (parentsToSave, variationsToSave);
         }
 
-        private static bool TryReserveProduct(string productId, ISet<string> alreadySavedIds, ISet<string> pendingIds)
+        private static bool TryReserveProduct(string productId, HashSet<string> alreadySavedIds, HashSet<string> pendingIds)
         {
             return !alreadySavedIds.Contains(productId) && pendingIds.Add(productId);
         }
@@ -777,7 +777,7 @@ namespace VirtoCommerce.CatalogModule.Data.ExportImport
 
         private static void DetachAssociations(
             CatalogProduct product,
-            IDictionary<string, IList<ProductAssociation>> associationBackupMap)
+            Dictionary<string, IList<ProductAssociation>> associationBackupMap)
         {
             if (!product.Associations.IsNullOrEmpty())
             {
@@ -799,7 +799,7 @@ namespace VirtoCommerce.CatalogModule.Data.ExportImport
         }
 
         private async Task SaveProductBatchAsync(
-            IList<CatalogProduct> products,
+            List<CatalogProduct> products,
             ImportStageContext stage,
             ProductImportContext context)
         {
@@ -826,7 +826,7 @@ namespace VirtoCommerce.CatalogModule.Data.ExportImport
         }
 
         private async Task ImportProductAssociationsAsync(
-            IReadOnlyDictionary<string, IList<ProductAssociation>> associationBackupMap,
+            Dictionary<string, IList<ProductAssociation>> associationBackupMap,
             ExportImportProgressInfo progressInfo,
             Action<ExportImportProgressInfo> progressCallback)
         {
--- a/src/VirtoCommerce.CatalogModule.Data/ExportImport/CatalogExportImport.cs
+++ b/src/VirtoCommerce.CatalogModule.Data/ExportImport/CatalogExportImport.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using Newtonsoft.Json;
@@ -103,21 +104,26 @@ namespace VirtoCommerce.CatalogModule.Data.ExportImport
             var progressInfo = new ExportImportProgressInfo { Description = "loading data..." };
             progressCallback(progressInfo);
 
-            using var sw = new StreamWriter(outStream);
-            using var writer = new JsonTextWriter(sw);
-            await writer.WriteStartObjectAsync(cancellationToken);
+            using var package = CatalogExportPackage.Create(outStream, options?.HandleBinaryData == true);
+            using (var streamWriter = new StreamWriter(package.CatalogStream, new UTF8Encoding(false), 1024, leaveOpen: true))
+            using (var writer = new JsonTextWriter(streamWriter))
+            {
+                await writer.WriteStartObjectAsync(cancellationToken);
 
-            await ExportPropertyGroupsAsync(writer, progressInfo, progressCallback, cancellationToken);
-            await ExportPropertiesAsync(writer, progressInfo, progressCallback, cancellationToken);
-            await ExportPropertyDictionaryItemsAsync(writer, progressInfo, progressCallback, cancellationToken);
-            await ExportCatalogsAsync(writer, progressInfo, progressCallback, cancellationToken);
-            await ExportCategoriesAsync(writer, options, progressInfo, progressCallback, cancellationToken);
-            await ExportProductsAsync(writer, options, progressInfo, progressCallback, cancellationToken);
-            await ExportProductConfigurationsAsync(writer, progressInfo, progressCallback, cancellationToken);
-            await ExportMeasuresAsync(writer, progressInfo, progressCallback, cancellationToken);
+                await ExportPropertyGroupsAsync(writer, progressInfo, progressCallback, cancellationToken);
+                await ExportPropertiesAsync(writer, progressInfo, progressCallback, cancellationToken);
+                await ExportPropertyDictionaryItemsAsync(writer, progressInfo, progressCallback, cancellationToken);
+                await ExportCatalogsAsync(writer, progressInfo, progressCallback, cancellationToken);
+                await ExportCategoriesAsync(writer, options, package, progressInfo, progressCallback, cancellationToken);
+                await ExportProductsAsync(writer, options, package, progressInfo, progressCallback, cancellationToken);
+                await ExportProductConfigurationsAsync(writer, progressInfo, progressCallback, cancellationToken);
+                await ExportMeasuresAsync(writer, progressInfo, progressCallback, cancellationToken);
 
-            await writer.WriteEndObjectAsync(cancellationToken);
-            await writer.FlushAsync(cancellationToken);
+                await writer.WriteEndObjectAsync(cancellationToken);
+                await writer.FlushAsync(cancellationToken);
+            }
+
+            await package.CompleteAsync(cancellationToken);
         }
 
         private async Task ExportPropertyGroupsAsync(JsonTextWriter writer, ExportImportProgressInfo progressInfo, Action<ExportImportProgressInfo> progressCallback, CancellationToken cancellationToken)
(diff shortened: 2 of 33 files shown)


src/VirtoCommerce.CatalogModule.Data/ExportImport/CatalogExportImport.cs
```diff
@@ -190,21 +208,28 @@ await writer.SerializeArrayWithPagingAsync(_jsonSerializer, _batchSize, async (s
...
             progressInfo.Description = "Categories exporting...";
             progressCallback(progressInfo);
 
+            var isCountQuery = true;
             await writer.WritePropertyNameAsync("Categories", cancellationToken);
             await writer.SerializeArrayWithPagingAsync(_jsonSerializer, _batchSize, async (skip, take) =>
             {
                 var searchResult = await _categorySearchService.SearchAsync(new CategorySearchCriteria { Skip = skip, Take = take });
-                LoadImages(searchResult.Results.OfType<IHasImages>().ToArray(), progressInfo, options.HandleBinaryData);
+                if (isCountQuery)
```
OlegoO: Remove  isCountQuery.

src/VirtoCommerce.CatalogModule.Data/ExportImport/CatalogExportImport.cs
```diff
@@ -213,20 +238,28 @@ await writer.SerializeArrayWithPagingAsync(_jsonSerializer, _batchSize, async (s
...
             progressInfo.Description = "Products exporting...";
             progressCallback(progressInfo);
 
+            var isCountQuery = true;
             await writer.WritePropertyNameAsync("Products", cancellationToken);
             await writer.SerializeArrayWithPagingAsync(_jsonSerializer, _batchSize, async (skip, take) =>
             {
                 var searchResult = await _productSearchService.SearchAsync(new ProductSearchCriteria { Skip = skip, Take = take, ResponseGroup = ItemResponseGroup.Full.ToString() });
-                LoadImages(searchResult.Results.OfType<IHasImages>().ToArray(), progressInfo, options.HandleBinaryData);
+                if (isCountQuery)
```
OlegoO: Remove isCountQuery.

src/VirtoCommerce.CatalogModule.Data/ExportImport/CatalogPackageFormat.cs
```diff
@@ -0,0 +1,82 @@
...
+    public const int Sha256HexLength = 64;
+    public const int CopyBufferSize = 81920;
+
+    // These are abuse-prevention ceilings, not operational catalog-size limits.
+    public const int MaximumEntryCount = 1_000_000;
+    public const long MaximumPackageLength = 1L << 40;
+
+    public static string CreateBinaryDataReference(string sourceUrl)
+    {
+        ArgumentException.ThrowIfNullOrEmpty(sourceUrl);
```
OlegoO, line 27: Replace with either sourceUrl or GUID

OlegoO (dismissed):

src/VirtoCommerce.CatalogModule.Data/ExportImport/CatalogExportImport.cs
ToxaKZ: Done

src/VirtoCommerce.CatalogModule.Data/ExportImport/CatalogExportImport.cs
ToxaKZ: Done

src/VirtoCommerce.CatalogModule.Data/ExportImport/CatalogPackageFormat.cs
ToxaKZ: Done

src/VirtoCommerce.CatalogModule.Data/ExportImport/CatalogExportImport.cs
```diff
@@ -868,51 +1045,192 @@ private static void ResetRedundantReferences(object entity)
...
+
+        private static IEnumerable<Asset> GetAssets<T>(IEnumerable<T> entities)
+        {
+            return new { entities }
+                .GetFlatObjectsListWithInterface<IHasAssets>()
+                .Where(x => !x.Assets.IsNullOrEmpty())
+                .SelectMany(x => x.Assets);
+        }
+
+        private static string GetRelativeUrl(AssetBase asset)
```
DmitryGrishinVirtoworks, line 1192: Do I understand correctly that GetRelativeUrl no longer falls back to Url when RelativeUrl is present but fails IsRelativeBlobUrl? The old code seemed to prefer a valid Url and only fall back to RelativeUrl. Is it intentional that an asset with a stale/invalid RelativeUrl now returns null even if Url itself would have been usable?

src/VirtoCommerce.CatalogModule.Data/ExportImport/CatalogExportImport.cs
```diff
@@ -868,51 +1045,192 @@ private static void ResetRedundantReferences(object entity)
...
+                if (handleBinaryData && !string.IsNullOrEmpty(sourceUrl))
+                {
+                    asset.Url = sourceUrl;
+                    await ExportBinaryDataAsync(asset, sourceUrl, package, cancellationToken);
                 }
             }
         }
 
-        private void ImportImages(IHasImages[] haveImagesObjects, ExportImportProgressInfo progressInfo)
+        private async Task ExportBinaryDataAsync(AssetBase asset, string sourceUrl, CatalogExportPackage package, CancellationToken cancellationToken)
```
DmitryGrishinVirtoworks, line 1081: It looks like ExportBinaryDataAsync no longer has access to ExportImportProgressInfo and only logs a warning on blob-read failure, whereas the old LoadImages also added it to progressInfo.Errors. Was dropping this from the errors list a deliberate choice? Should something be added so callers relying on progressInfo.Errors can still detect that some assets were skipped?

src/VirtoCommerce.CatalogModule.Data/ExportImport/CatalogExportImport.cs
```diff
@@ -868,51 +1045,192 @@ private static void ResetRedundantReferences(object entity)
...
-                    {
-                        progressInfo.Errors.Add(ex.Message);
-                    }
+                    await ExportBinaryDataAsync(image, sourceUrl, package, cancellationToken);
+                }
+            }
+
+            foreach (var asset in GetAssets(entities))
+            {
+                var sourceUrl = GetRelativeUrl(asset);
```
DmitryGrishinVirtoworks, line 1069: Am I reading this right that image.Url gets normalized to the relative source URL unconditionally, but asset.Url only gets normalized when handleBinaryData is true? Is this asymmetry intentional, or should both follow the same rule?

src/VirtoCommerce.CatalogModule.Data/ExportImport/CatalogExportImport.cs
```diff
@@ -95,29 +111,39 @@ private async Task LoadSettingsAsync()
             }
         }
 
-        public async Task DoExportAsync(Stream outStream, ExportImportOptions options, Action<ExportImportProgressInfo> progressCallback, CancellationToken cancellationToken)
+        public Task DoExportAsync(Stream outStream, ExportImportOptions options, Action<ExportImportProgressInfo> progressCallback, CancellationToken cancellationToken)
```
DmitryGrishinVirtoworks, line 114: Just want to confirm - the legacy DoExportAsync overload (kept for IExportSupport compatibility) now seems to write a ZIP archive to outStream when HandleBinaryData=true, whereas before it always wrote plain JSON regardless of that flag? If so, could this break existing callers that still expect JSON from this overload?

src/VirtoCommerce.CatalogModule.Data/ExportImport/CatalogImportPackage.cs
```diff
@@ -0,0 +1,334 @@
...
+            var signature = new byte[CatalogPackageFormat.SignatureLength];
+            var signatureLength = await ReadSignatureAsync(inputStream, signature, cancellationToken);
+
+            if (!IsZipArchive(signature, signatureLength))
+            {
+                var catalogStream = new PrefixReadStream(signature, signatureLength, inputStream, leaveOpen: true);
+                return new CatalogImportPackage(inputStream, catalogStream, binaryDataReader);
+            }
+
+            temporaryStream = TemporaryFileStream.Create();
```
DmitryGrishinVirtoworks, line 67: Is it intentional that OpenInternalAsync always copies the whole input into a temp file before opening the ZipArchive, even when inputStream is already seekable? Would it make sense to skip the copy in that case, or is there a reason (validation, uniform code path, etc.) it's always done?

src/VirtoCommerce.CatalogModule.Data/ExportImport/CatalogExportImport.cs
```diff
@@ -868,51 +1045,192 @@ private static void ResetRedundantReferences(object entity)
...
+        {
+            if (!string.IsNullOrWhiteSpace(asset.RelativeUrl))
+            {
+                return IsRelativeBlobUrl(asset.RelativeUrl) ? asset.RelativeUrl : null;
+            }
+
+            return IsRelativeBlobUrl(asset.Url) ? asset.Url : null;
+        }
+
+        private static bool IsRelativeBlobUrl(string url)
```
DmitryGrishinVirtoworks, line 1202: Was there a reason to implement IsRelativeBlobUrl from scratch here instead of reusing the existing IsAbsoluteUrl() extension from Platform.Core that this same file used to rely on? Could the two behave differently for some URL shapes?

src/VirtoCommerce.CatalogModule.Core/Model/AssetBase.cs
```diff
@@ -19,6 +19,12 @@ protected AssetBase(string typeName)
         }
         public string RelativeUrl { get; set; }
         public string Url { get; set; }
+        /// <summary>
+        /// Gets or sets the transient relative reference to binary data in a catalog backup package.
+        /// The property is populated only while a backup is exported or restored and is not persisted.
+        /// </summary>
+        [JsonProperty(NullValueHandling = NullValueHandling.Ignore)]
+        public string BinaryDataReference { get; set; }
```
DmitryGrishinVirtoworks, line 27: Just checking - does adding a public settable BinaryDataReference to AssetBase mean it now shows up in the regular product/category REST payloads too, not just export/import? Is that expected?

src/VirtoCommerce.CatalogModule.Data/ExportImport/CatalogExportPackage.cs
```diff
@@ -0,0 +1,149 @@
...
+
+            var manifestEntry = _archive.CreateEntry(CatalogPackageFormat.ManifestEntryName, CompressionLevel.NoCompression);
+            await using var manifestStream = manifestEntry.Open();
+            await manifestStream.WriteAsync(CatalogPackageFormat.CreateManifest(), cancellationToken);
+        }
+
+        _isCompleted = true;
+    }
+
+    public void Dispose()
```
DmitryGrishinVirtoworks, line 138: If an exception or cancellation happens before CompleteAsync runs, does Dispose() still finalize the ZIP archive on outStream? Could that leave a structurally valid but incomplete package (missing catalog.json/package.json) behind, and is that a concern for callers who might not check completion status?

src/VirtoCommerce.CatalogModule.Data/ExportImport/CatalogExportImport.cs
ToxaKZ: Yes, this is intentional. `RelativeUrl` is treated as the authoritative storage path when it is present. If it is external or invalid, we fail closed instead of falling back to a different `Url` value, because that fallback could bypass the external/unsafe URL protection. `DoExportAsync_RelativeUrlWinsAndAbsoluteSchemesAreSkipped` covers both preferring a valid relative URL and not falling back for `s3`, `https`, or scheme-relative URLs.

src/VirtoCommerce.CatalogModule.Data/ExportImport/CatalogExportImport.cs
ToxaKZ: Yes, dropping this from `progressInfo.Errors` was deliberate. BackupRestore treats entries in `Errors` as module export failures and reports `Exported ... with errors`. A missing physical blob is non-fatal here: the catalog entity is still exported, its `BinaryDataReference` is omitted, remaining files continue exporting, and the condition is logged at Warning level. The test verifies that the warning is logged while `progressInfo.Errors` stays empty.

src/VirtoCommerce.CatalogModule.Data/ExportImport/CatalogExportImport.cs
ToxaKZ: The asymmetry is intentional for backward compatibility. Images were historically normalized to `RelativeUrl` during export regardless of `HandleBinaryData`, while file assets were not modified by the old export path. Assets are normalized when binary packaging is enabled because their `Url` must match the sidecar source/destination path. With binary handling disabled, the raw JSON keeps the previous asset shape.

src/VirtoCommerce.CatalogModule.Data/ExportImport/CatalogExportImport.cs
ToxaKZ: Yes, this is intentional. The legacy single-stream `IExportSupport` contract has no binary sidecar writer, so with `HandleBinaryData=true` it uses a versioned nested package to avoid inline base64 and full binary buffering. With binary handling disabled it still emits raw JSON. Import detects JSON versus ZIP by signature and continues to accept legacy plain JSON/base64 packages. The new BackupRestore flow uses the sidecar-aware overload and receives readable JSON plus top-level assets.

src/VirtoCommerce.CatalogModule.Data/ExportImport/CatalogImportPackage.cs
ToxaKZ: Yes, the disk spool is intentional. Nested packages can arrive through non-seekable streams, and relying on `ZipArchive` there can introduce hidden in-memory buffering. Spooling gives us bounded memory usage and one validation/ownership path for both stream types. The normal sidecar-aware format is raw JSON and does not take this ZIP-spooling path; this applies to the legacy nested package. A seekable-only optimization is possible, but it would add a second lifetime path for the legacy format.

src/VirtoCommerce.CatalogModule.Data/ExportImport/CatalogExportImport.cs
ToxaKZ: The difference is intentional. Platform `IsAbsoluteUrl()` recognizes only a whitelist of schemes. For a blob/package path we must reject every absolute URI, including custom schemes such as `s3`, as well as scheme-relative URLs starting with `//`. Reusing `IsAbsoluteUrl()` would allow an unrecognized absolute scheme to be treated as a relative blob URL.

src/VirtoCommerce.CatalogModule.Core/Model/AssetBase.cs
ToxaKZ: Yes, it is an additive public model property and can be visible in the API schema. It is transient, not persisted, populated only by the export/import pipeline, and regular models loaded from persistence leave it null; `NullValueHandling.Ignore` ensures it is omitted from normal JSON payloads. This follows the existing transient `BinaryData` pattern on `Image` and `Asset`. Hiding it completely from the API model while retaining export/import serialization would require a separate export DTO or custom serializer/converter.

src/VirtoCommerce.CatalogModule.Data/ExportImport/CatalogExportPackage.cs
ToxaKZ: Yes, `Dispose()` finalizes the archive central directory. If export throws or is cancelled, however, the output stream is invalid and the caller must discard it, just as it would discard partially written JSON; bytes already written to a non-seekable stream cannot be rolled back. Import validates both `package.json` and `catalog.json` before saving entities, so a structurally readable but incomplete package is rejected. Disposing the archive is still required to release resources without masking the original failure.

DmitryGrishinVirtoworks (approved):

